PR #15, opened 2017-07-26
Submit script

jithinbp:
There was a syntax error in views.py which caused the CI to fail for python3, and as a result the web-link for the app is also not working.

Heroku has now been set to wait for CI to pass before auto-deploying.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,5 @@
-<<<<<<< HEAD
 [Heroku App](https://pslab-stage.herokuapp.com/)
-=======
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
```
viveksb007: Need to remove these extra characters after solving conflicts.

jithinbp:
@viveksb007 Fixed. Thank you

PR #17, opened 2017-07-26
Backend routine for submitting scripts

### LINKED ISSUE #16: Backend functionality for storing user created scripts
@viveksb007 @asitava1998 @CloudyPadmal In the [last PR](https://github.com/fossasia/pslab-remote/pull/14) , I embedded ace.js text editor on a page that allowed users to write their own Python code.

Now the question is how to store these.
Should they go into a database along with the Primary key user-email and a unique id for searchability, 
or
Should they be stored as text files in a directory.

Which begs the larger question ... if users are allowed to upload experiment results/screenshots as the app evolves, where is such data typically stored?

jithinbp:
scripts go into a new table along with the user-id for searchability and access control

fixes #16 .

### DIFF
--- a/app/views.py
+++ b/app/views.py
@@ -80,6 +80,14 @@ def logout():
     session.pop('user',None)
     return redirect('/')
 
+@app.route('/showAddScript')
+def showAddScript():
+	if session.get('user'):
+		print session['user']
+		return render_template('addScript.html',author = session['user'][0])
+	else:
+		return render_template('error.html',error = 'Unauthorized Access')    
+
 
 
 """
--- a/app/views.py
+++ b/app/views.py
@@ -109,7 +109,6 @@ def addScript():
 		else:
 			return render_template('error.html',error = 'Unauthorized Access')
 	except Exception as e:
-		print e
 		return render_template('error.html',error = str(e))
 
 
--- a/app/views.py
+++ b/app/views.py
@@ -83,7 +83,7 @@ def logout():
 @app.route('/showAddScript')
 def showAddScript():
 	if session.get('user'):
-		print session['user']
+		#print (session['user'])
 		return render_template('addScript.html',author = session['user'][0])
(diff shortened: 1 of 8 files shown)


app/views.py
```diff
@@ -83,13 +83,38 @@ def logout():
...
 	else:
 		return render_template('error.html',error = 'Unauthorized Access')    
 
 
 
+@app.route('/addScript',methods=['POST'])
+def addScript():
+	try:
+		if session.get('user'):
+			_user = session.get('user')[1]
```
viveksb007: Could you also give ```_user``` as a parameter to addScript? Then in future we can make a call from Android or Desktop app to save script.

app/views.py
```diff
@@ -83,13 +83,38 @@ def logout():
...
+				db.session.add(newSnippet)
+				db.session.commit()
+				return redirect('/userHome')
+			except Exception as exc:
+				return render_template('error.html',error = 'Write Failed.') 
+
+		else:
+			return render_template('error.html',error = 'Unauthorized Access')
+	except Exception as e:
+		print e
```
viveksb007: This is causing build error.

app/views.py
jithinbp: The _user variable is automatically obtained from the session variable
`_user = session.get('user')[1] `
It is stored there when the user logs in, and if a login event has not occurred, an unauthorized access error will be raised. Did not make sense to allow a separate user field if the login event is necessary anyway. Is that okay?

app/views.py
jithinbp: fixed

app/views.py
viveksb007: Yeah was just saying that this restricts the use of this function to only remote app. If we send user parameter in function we can call this endpoint from any app. But android app hasn't developed to the point to save scripts, so we can change things later.

PR #21, opened 2017-07-29
added frontend code for displaying user submitted scripts

### LINKED ISSUE #20: Create frontend call for fetching and displaying list of scripts in the user's home page
+ The ability to write and submit scripts has already been implemented
+ Now we need to show the list of submitted scripts on the user's homepage

+ Further down the line(separate issue), the user should be able to view, edit, delete or execute these scripts.

jithinbp:
fixes #20 .

psl_virt_home

----

Uses jquery templates for parsing and displaying json.
A small script counter has also been added which is shown next to the title


mariobehling:
You are starting to develop the backend and front-end in one app. Earlier I had the suggestion to implement the application in a way that it communicates through an API and that you could run a front-end client with it. We are following this approach with Open Event. It is more sustainable and easier to maintain.

jithinbp:
@mariobehling I'm still trying to understand the open-event structure, but it's a very advanced software now.  I have gathered the following
+ https://github.com/fossasia/open-event-orga-server is just the pure API server. various apps can communicate with this
+ https://github.com/fossasia/open-event-webapp appears to be standalone
+ https://github.com/fossasia/open-event-orga-server/issues/3457 . issue where niranjan has given detailed comments.

The last comment thread has provided a lot of insight, and the frontend was moved a separate repository by the end of the discussion. But this frontend does not rely on flask, and uses ember-cli to serve pages.

Current structure I'm following for pslab-remote
+ frontend
  + app/templates : directory that stores frontend code which is mostly html + js
  + app/static : directory that stores css , js, font  files

+ backend
  + views.py which contains the API calls . These include SignIn, SignOut, GetScriptList etc which either return JSON serialized objects, or simply provide a passthrough to the html files stored in app/templates.
  + modification of frontend files( stored in app/templates) by the backend only occurs in the following instances
    + error.html : The backend includes a custom error message into it before serving the page.
    + `render_template('addScript.html',author = session['user'][0])` where the user id is passed but is not really necessary , and can be removed.

Can you add someone into the conversation who can suggest how to best separate the view from the API?
I believe the error page can also be made static , and an AJAX call to retrieve and display the error message can be built into it. And then we can more or less agree that the frontend and backend is stored in separate directories. However, we will still need flask to act as a passthrough for static pages unless someone can chip in with a method to instruct the web-server to directly serve files contained in app/templates

Excerpt from another [thread](https://teamtreehouse.com/community/is-there-a-different-type-of-approach-to-back-ends-in-python-than-flask-django-etc) I went through .
```
Flask, like all frameworks(the ones I know) use the MVC architecture. Look into that as it is a must for any developer, especially a web developer. The short explanation, the V in MVC is the VIEW, html in your case. This is what the user will see, the M stands for MODEL, this is some sort of storage or values. Your VIEW will obviously want to display some of that data but can't get access to them. For example, you can't get data from a database like MySQL with HTML code, thats where the CONTROLLER comes into play. It is the intermediary between the two.
```

thank you

In the meantime, I've made a separate branch where I have moved the html files in the templates folder to the static folder, and will be removing all render_template calls from views.py . 
Authentications and redirections to be handled via AJAX, JS, and the only shared segment will be 3-4 lines of code instructing flask to serve static files from the frontend's directory.

niranjan94:
@jithinbp I think you mis-understood the structure of the Open Event application (about the frontend).

- [API Server - fossasia/open-event-orga-server](https://github.com/fossasia/open-event-orga-server) - Pure API server. No UI code. Strictly compatible with the [json-api](http://jsonapi.org) specification.
- [Frontend - fossasia/open-event-frontend](https://github.com/fossasia/open-event-frontend) - A completely decoupled frontend. Uses Ember.js framework. 

So, what I'm trying to say is, 

- Don't mix up the frontend and the backend code in anyway. Either have two different repositories, or in the same repository, have it like this

	```
	The Repository
	├── frontend
	│   └── The root of the frontend project.
	└── api-server
	    └── The root of the API Server
	```
- Ensure your API server follows a standard API specification such as [json-api](http://jsonapi.org) from the start. This will be very useful once you start having a lot of clients consuming the API.

jithinbp:
@niranjan94 Thanks for your input. I understood that the frontend was completely decoupled, but am not familiar with ember yet.

In pslab-remote , most of the frontend is decoupled because they reside in the templates directory, but it was not an ideal scenario because some of the API calls responded with redirections to an error page if necessary instead of json data with a false flag set or something.

In the meantime, I have tried to fully decouple these two in a separate [branch](https://github.com/jithinbp/pslab-remote/tree/separate_api/), and all the frontend code resides in the `static` folder now. 
However, I am still using the same flask server to host both the API calls ( They do not conform to json-api yet, but I'll take care of that once this is stabilised), as well as these static html files , so there's still a small section of code in the backend that redirects all file requests to the static folder.
+ backend app :  [views.py](https://github.com/jithinbp/pslab-remote/blob/separate_api/app/views.py)
+ frontend stuff : [app/static](https://github.com/jithinbp/pslab-remote/tree/separate_api/app/static) 

In the current scenario, it does not take advantage of a framework such as ember, and all data-validations and redirections are handled inside AJAX calls [ e.g., if a user submits a code snippet, the call will determine if the title is unique, and accordingly the page may redirect to the user's home page, or show an error message]. I will probably trash this branch once I find a more scalable configuration.

Once you've found time to take a look at the branch linked above, what path do you suggest I take?
+ Rewrite the frontend code using the ember.js framework. It will be time consuming for me to start afresh with this, but I shall try if that makes it scalable.
+ make two separate flask apps to serve frontend&backend. I don't know if this is acceptable, but I suppose I can also make one single app treat all calls to /api/ as API requests, and the rest as static files.

niranjan94:
- Don't use branches in same repo to hold completely different stuff. (like frontend and backend)
- You don't need to use Ember. It is just an option. But, Why is flask needed for a frontend ? Flask and python are for backend.

jithinbp:
I'm not using branches to hold different things.
I simply made a branch that modifies the whole thing so that frontend is fully decoupled. The branch part was only to ensure I wouldn't end up destroying the whole thing.

In that branch, It should suffice to launch index.html from the static folder separately instead of going through flask. I'm not sure how to instruct Heroku to display a static html file after launching the backend, but it must be possible, and I'll get to that later.

It looks like I'm going to have to use a framework such as Ember because links such as the `logout` button must also go through an AJAX call (Correct me if there's a better way),  and that'll make for some large and messy code.

Thanks for your suggestions. I'll continue work on the new branch to see how it shapes up, and ignore this PR for the time being.

PR #49, opened 2017-08-12
added codeEvaluator to execute code snippets extracted from the db.

### LINKED ISSUE #44: Create backend API method and handler for evaluation of python programs
+ handler will execute remotely submitted code in a sandbox
+ stdout will be redirected to formatted html
+ results will be displayed in the webapp

### LINKED ISSUE #47: Ember controller and template for executing scripts remotely and displaying results
+ send script ID to API server for execution.
+ Open modal and display result/error.

### LINKED ISSUE #48: Add a 'back' button in the `add script` page of the webapp
The webapp only has an option to logout as of now.

jithinbp:
Screencast : A python program to make an array of 30 equidistant points from 0 to 2*pi , and output the array as well as sine(array) is run on the backend and results are displayed in the webapp. The same can be used for all sorts of PSLab functions such as capture1, get_capacitance etc

psl_virt_runscript

+ Added modal to user-home.hbs to submit code and view the results
+ Created a controller to submit code-id to the API and process the returned results
+ linked the controller function to the execute icon against each script in the list on the user's homepage
+ Created API method to fetch a script from the database and pass it to codeEvaluate.py which will run the code in a sandboxed environment that includes all the functions of PSLab, as well as libraries such as numpy
+ Wrote codeEvaluator.py which runs scripts and returns the console output.

fixes #44 (backend API method + evaluator ).
fixes #47 (EmberJS controller and template for remote code execution
fixes #48 : Add a 'back' button in the `add script` page of the webapp

### DIFF
--- a/app/codeEvaluator.py
+++ b/app/codeEvaluator.py
@@ -1,0 +1,47 @@
+#from cStringIO import StringIO
+from io import BytesIO
+import sys
+
+import numpy as np
+from flask import json
+from collections import OrderedDict
+
+class Evaluator:
+	def __init__(self,functionList):		
+		self.html = ''
+		self.hasPlot=False
+
+		self.evalGlobals={}
+		self.functionList = functionList
+		self.evalGlobals = functionList.copy()
+		self.evalGlobals['print_']=self.printer
+		
+	
+	def printer(self,txt):
+		self.html+='''<div id="print-statement" class="row well well-sm">%s</div>'''%txt
+
+	def runCode(self,code):
+		old_stdout = sys.stdout
+		sys.stdout = results = BytesIO()
+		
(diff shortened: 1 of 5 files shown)


jithinbp:
@viveksb007 It's part of the roadmap I have in mind.  If you like, I can add a tab to show the source code.
In the example I have shown the user executing his own script, but in the future there will be scripts common to all users with execute permissions, not edit.

Further down the line I plan to add features such as loops which will basically contain JS code. e.g., if you want a continuously running oscilloscope, this approach will not work because the python script won't finish executing. Therefore, the script has to finish and render web content that will then utilise the API call to execute single functions, and act accordingly. I already made some prototypes, but Niranjan might have other suggestions/plans.

PR #55, opened 2017-08-15
Json structured script outputs

jithinbp:
The API server returns a JSON object containing the output of a script retrieved via the submitted id. Previously, a string containing the console output was being returned, but in the interest of providing greater flexibility, a JSON object containing a list of various sections of the output will now be the return value.
 This will also ensure that no HTML/JS needs to be generated by the API, and all formatting will handled by the Ember App based on properties defined in the JSON object.

```
e.g. : print ('hello world') will be output as
{'status':True,
'Filename':script.title,
'Date':script.pub_date
'Id':script.id,
'result':[{"name":"print","type":"span","value":"hello world"}],
}
```
The various types implemented in the boilerplate code that this PR provides, includes 'span', and 'span-well'

------
+ example.py
```
print ('testing')
print ('testing some changes..... ')
print_('highlighted print statement')
```
+ screenshot of output of example.py (id=4 on my testing server)
psl_virt_json

+ JSON returned by the API [localhost:8000/runScriptById , POST: id=4 (example.py)] from which the above was generated
```
{"Date": "Tue, 01 Aug 2017 21:39:12 GMT", "Filename": "example.py", "Id": 4,
 "result": [
  {"name": "print", "type": "span", "value": "('testing',)"},
  {"name": "print", "type": "span", "value": "('testing some changes..... ',)"},
  {"class": "row well", "name": "print", "type": "span", "value": "highlighted print statement"}
  ],
"status": true}
```


frontend/app/templates/user-home.hbs
```diff
@@ -174,14 +174,30 @@
...
+            
+						{{#each codeResults as |element|}}
+							{{#if (eq element.type 'input')}}
+								<label>{{element.label}}</label>
+								{{input id=element.name}}
+							{{/if}}
+							{{#if (eq element.type 'button')}}
+								<button id="{{element.name}}" {{action 'runAction' element.action}}>{{element.label}}</button>
+							{{/if}}
+							{{#if (eq element.type 'span')}}
```
niranjan94: Call it text. Because ... this is just a text right ...

frontend/app/helpers/eq.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
```
niranjan94: Remove this helper. And instead of this, add this addon https://github.com/jmurphyau/ember-truth-helpers

frontend/tests/integration/helpers/eq-test.js
```diff
@@ -0,0 +1,17 @@
+
```
niranjan94: remove this too

PR #67, opened 2017-08-20
added plot setData target for buttons

### LINKED ISSUE #66: Enable routing button callback results to plots
+ At present, buttons when clicked can get numeric/string responses from an API and update a label
+ Now they should be able to fetch arrays and update plot objects

jithinbp:
+ enabled routing button callback results to plots
+ fixed a bug in the toUnique method

```
import numpy as np
x=np.linspace(0,2*np.pi,30)
plt = plot(x, np.sin(x))
button('get voltage',"get_voltage('CH1')")
button('update plot',"capture1('CH1',100,10)","update-plot",target=plt)
```
psl_virt_button_plot
As you can see, clicking the button executes its target and loads the data, but since we have not yet implemented conversion of x and y datasets into x,y pairs, it makes two separate plots for each axis

What more needs to be done:
+ The plots accept x,y datapoint pairs. But since PSLab capture methods return xdata, ydata ,  transpose([xdata,ydata]) needs to be carried out in the frontend

fixes #66


frontend/app/helpers/column-stack.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
```
niranjan94: I think this can be removed.

jithinbp: Is there a case for putting package-lock.json in the gitignore?
I didn't manually change it. After you added eslint dependencies, I got missing dep errors, so I ran 'npm install' , which is probably what changed it

niranjan94: `package-lock.json` is meant to be committed. Which version of npm are you on ?

jithinbp: npm 5.3.0
I'll restore its contents from the master branch then. although I'm not sure what I'm doing here

niranjan94: Weird. Its okay. don't revert. Lets merge.

jithinbp: okay, cool. Travis just finished.

PR #69, opened 2017-08-22
added underscore dependency and enabled array manipulations

### LINKED ISSUE #68: Add underscore.js dependency
+ Will be required for array manipulations and other mathematical functions
+ First use will be to convert separat xdata and ydata to xy pairs.

jithinbp:
+ Added ember-cli-underscore
+ controllers/user-home.js checks the 'stacking' argument of data passed to the plot, and makes a transpose of [xdata,ydata] if necessary
+ fixed some pending eslint warnings

psl_virt_plot_array

---
small change I'm unable to make : Can't call the replot method on an existing plot
```
// alert(JSON.stringify($(`#${actionDefinition.success.target}`).data(), null, 4));
// $(`#${actionDefinition.success.target}`).data('jqplot').replot({ data: [_.zip.apply(null, resultValue)] });
$.jqplot(actionDefinition.success.target, [_.zip.apply(null, resultValue)]).replot();
```
Uncomment the alert to see that data attached to the jquery selector is {}  even though the plot was created in that div by a component. The selector works fine with a regular small js snippet, and according to the [docs](http://www.jqplot.com/deploy/dist/examples/selectorSyntax.html), it is the correct way to update a plot. @niranjan94  any suggestions on why there seem to be no methods attached to the selector (I've cross checked with chrome's inspector) ?
Another way is to store a global variable referencing the plot, but I would prefer the selector.

fixed #68


frontend/package.json
```diff
@@ -36,6 +36,7 @@
     "ember-cli-surge": "^1.1.20",
     "ember-cli-template-lint": "^0.7.3",
     "ember-cli-uglify": "^1.2.0",
+    "ember-cli-underscore": "0.0.3",
```
niranjan94: Use lodash instead please .. 
https://github.com/mike-north/ember-lodash 

Has the same functions as underscore and more and is better performance wise.

jithinbp: okay, no issue.
 What about the selector problem? If it because the plot was created in a  component? Do I have to place the replot method also in the same component and reference it somehow? thanks

PR #72, opened 2017-08-22
Array manipulations for processing data from capture calls

### LINKED ISSUE #68: Add underscore.js dependency

### LINKED ISSUE #70: Second plot in user script replaces the first if name is not specified
how to reproduce the bug
```python
import numpy as np
x=np.linspace(0,2*np.pi,30)
plt = plot(x, np.sin(x))
plt2 = plot(x, np.sin(x))
```
right way
```python
import numpy as np
x=np.linspace(0,2*np.pi,30)
plt = plot(x, np.sin(x))
plt2 = plot(x, np.sin(x),name = 'newplotid')
```
default name is `myPlot` . However, the toUnique method is not being invoked , and hence the same name is used for all plots.

### LINKED ISSUE #71: user-home.js/update-plot : add different array input types. Support  capture 2
+ the stacking variable defaults to 'xy' which takes [xdata,ydata] and makes [[x1,y1],[x2,y2]... ] 
+ Extrapolate this for capture2 method
+ Extrapolate this for capture4 method

jithinbp:
+ Added ember-lodash . fixes #68 
+ controllers/user-home.js checks the 'stacking' argument of data passed to the plot, and makes a transpose of [xdata,ydata..] if necessary. fixes #70 
+ fixed some pending eslint warnings
+ made plot names unique. fixes #70 
+ added capture4 method to the dummy class

```python
import numpy as np
x=np.linspace(0,2*np.pi,30)
plt = plot(x, np.sin(x))
button('capture 1',"capture1('CH1',100,10)","update-plot",target=plt)

plt2 = plot(x, np.sin(x))
button('capture 2',"capture2(50,10)","update-plot",target=plt2,stacking='xyy')

plt3 = plot(x, np.sin(x))
button('capture 4',"capture4(50,10)","update-plot",target=plt3,stacking='xyyyy')
```
gifrecord_2017-08-22_125031

---
small change I'm unable to make : Can't call the replot method on an existing plot
```
// alert(JSON.stringify($(`#${actionDefinition.success.target}`).data(), null, 4));
// $(`#${actionDefinition.success.target}`).data('jqplot').replot({ data: [_.zip.apply(null, resultValue)] });
$.jqplot(actionDefinition.success.target, [_.zip.apply(null, resultValue)]).replot();
```
Uncomment the alert to see that data attached to the jquery selector is {}  even though the plot was created in that div by a component. The selector works fine with a regular small js snippet, and according to the [docs](http://www.jqplot.com/deploy/dist/examples/selectorSyntax.html), it is the correct way to update a plot. @niranjan94  any suggestions on why there seem to be no methods attached to the selector (I've cross checked with chrome's inspector) ?
Another way is to store a global variable referencing the plot, but I would prefer the selector.

fixed #68 ,fixed #71  , fixed #70


frontend/app/controllers/user-home.js
```diff
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import _ from 'lodash';
```
niranjan94: Import only what is needed. 

For example ... 
```
import { zip } from 'lodash'
```

frontend/app/controllers/user-home.js
```diff
@@ -170,7 +171,17 @@ export default Controller.extend({
             } else if (actionDefinition.success.type === 'display') {
               $(`#${actionDefinition.success.target}`).text(resultValue);
             } else if (actionDefinition.success.type === 'update-plot') {
-              $.jqplot(actionDefinition.success.target, resultValue).replot();
+              if (actionDefinition.success.stacking === 'xy') {
+                // alert(JSON.stringify($(`#${actionDefinition.success.target}`).data(), null, 4));
+                // $(`#${actionDefinition.success.target}`).data('jqplot').replot({ data: [_.zip.apply(null, resultValue)] });
+                $.jqplot(actionDefinition.success.target, [_.zip.apply(null, resultValue)]).replot();
```
niranjan94: Why do you need to use `Function.prototype.apply` when you could directly call the function ? What's the reason ?

frontend/bower.json
```diff
@@ -16,6 +16,7 @@
   ],
   "dependencies": {
     "bootstrap": "^3.3.7",
-    "jqplot": "*"
+    "jqplot": "*",
+    "underscore": "^1.8.3"
```
niranjan94: Remove this please.

frontend/app/controllers/user-home.js
niranjan94: And doesn't this work ? Instead of using the data ?

frontend/app/controllers/user-home.js
jithinbp: I had done this initially, but then there was an esllint error with using zip.appy . It asked me to use spread, and I couldn't figure out the syntax. So i reverted.

frontend/app/controllers/user-home.js
jithinbp: Could you please tell me the correct syntax? None of the combinations I tried after googling worked. Got a blank plot.
How do I enable console.log?

jithinbp:
@niranjan94 could you take a look and merge this now? thanks

PR #82, opened 2018-04-02
Show password checkbox to toggle password visibility in login and sig…

Vikum94:
A user can use the check-box to toggle password visibility in both Sign-up and login forms.


CloudyPadmal:
Travis build is giving some errors!

PR #126, opened 2026-08-15
feat: data model, migrations & state machine

ViRUS-0-0:
## Closes #120

> [!NOTE] 
> - This is a Stacked PR and will be ready to review after #121 gets merged

## Description
This PR introduces the foundational relational schema and the workflow state machine for the VEditor project. In accordance with the Phase 1 roadmap, the models establish the strict relational bounds of our service (`Event`, `Client`, `Talk`, `Job`, `Review`) while ensuring no media files are directly managed via SQLAlchemy. 

Additionally, this PR implements the core pipeline state machine as a pure Python dictionary mapping—explicitly avoiding heavy state-machine libraries and keeping the logic completely decoupled from database sessions and the FastAPI framework.

## Reviewer's Guide
To understand the flow of this PR, I recommend reviewing the files in the following order:

1. **`app/models.py` & `app/schemas.py`**
   - **What to look for:** Check the core SQLAlchemy 2.0 ORM mappings. Notice the use of PostgreSQL's `ARRAY` for `Client.event_ids` to avoid unnecessary join tables. 
   - **Schemas:** Review the Pydantic schemas which mirror the DB models and are configured with `from_attributes=True` for seamless future API serialization.
   
2. **`app/states.py`**
   - **What to look for:** This is the heart of the pipeline constraints. Notice that it has **zero external imports** from FastAPI, Celery, or `db.py`. 
   - **Logic:** Review the `TRANSITIONS` dictionary mapping and the `advance()` function. It takes a `Talk` object, validates the target state, and throws our custom, typed `InvalidTransitionError` upon illegal requests.

3. **`migrations/env.py` & `migrations/versions/0001_initial_migration.py`**
   - **What to look for:** We modified `env.py` to auto-generate sequential, Django-style migration IDs (`0001`, `0002`) instead of random hex strings by parsing the local `versions/` directory.

4. **`tests/test_states.py` & `tests/test_models.py`**
   - **What to look for:** Ensure you are satisfied with the test coverage. The state tests dynamically verify all legal graph transitions and explicitly cover critical human-gate paths (`pending_approval -> rejected` and `preview -> needs_work -> cutting`). The model tests run against a live Postgres DB to verify foreign key integrity and constraints.

## Acceptance Criteria Met
- [x] Initial Alembic migration generates all five tables cleanly.
- [x] Foreign key constraints appropriately map Jobs and Reviews back to Talks, and Talks back to Events.
- [x] `alembic upgrade head` runs cleanly natively and within Compose.
[...]

PR #127, opened 2026-08-16
feat: auth, bootstrap cli & ops view

### LINKED ISSUE #122: [Phase 1] Auth, bootstrap CLI & ops view
## Description
Everything needed to create a client, authenticate, and inspect talks/jobs before the public API lands in Phase 4. Auth is the security boundary the bootstrap CLI and ops view both sit behind, so it's built first within this issue and reused by the other two rather than each rolling its own check.

## Tasks
**Auth**
* [ ] app/auth.py — API key hashing (never store plaintext), a FastAPI dependency that reads X-API-Key, resolves it to a Client, and returns 401 if the key is missing or invalid
* [ ] Event-scope enforcement: the dependency checks the requested event_id is in Client.event_ids, returning 403 if not
* [ ] Ensure hashed keys and any key-derived values never appear in logs, error responses, or exception tracebacks
* [ ] Tests: missing key → 401, malformed/unknown key → 401, valid key + in-scope event → success, valid key + out-of-scope event → 403

⠀**Bootstrap CLI**
* [ ] veditor admin create-client command — creates an Event and a Client scoped to it, prints the raw API key exactly once (never persisted or logged in plaintext)
* [ ] Decide and document CLI signature: does it create the Event inline (e.g. --event-name "...") or require an existing --event-id? Original roadmap example (veditor admin create-client --event …) implies the latter — confirm and note the decision in the command's --help text
* [ ] Validate inputs (e.g. reject creating a Client with zero scoped events)
* [ ] Test: running the command produces a Client row, a hashed key in the DB, and a printed raw key that authenticates successfully against the auth dependency above

⠀**Ops view**
* [ ] GET /ops/talks — lists talks with current state, across events an operator's key can see
* [ ] GET /ops/talks/{id} — talk detail including its associated jobs
* [ ] GET /ops/jobs/{id} — job status and log_path
* [ ] All ops routes require a valid X-API-Key (reuse the auth dependency above — no separate auth mechanism)
* [ ] Keep these routes out of any documentation intended for external clients (internal/ops-only, revisit in Phase 7 if a real admin UI is warranted)
* [ ] Response payloads use the Pydantic schemas from #2, not raw ORM objects

## Acceptance Criteria
* [ ] No route that touches Talk/Job/Review data can be reached without passing through the auth dependency
* [ ] A client scoped to Event A cannot read or act on data belonging to Event B, even with an otherwise-valid key
* [ ] Raw API keys are never persisted — only their hash
* [ ] CLI creates Event/Client; the printed key authenticates successfully on first use
* [ ] Command behavior (Event creation vs. reference) is documented, not just implied by argument names
* [ ] An operator with a valid key can view talks/jobs across the events that key is scoped to
* [ ] Ops endpoints return real DB data via the Pydantic schemas, not raw ORM dumps

ViRUS-0-0:
# Closes #122 

> [!NOTE] 
> - This is a Stacked PR and will be ready to review after #126 gets merged

## Description
The easiest way to prove the branch works is to run the new test file we just added. This will run through all the authorized and unauthorized scenarios automatically.

Run this in your terminal:
```bash
uv run pytest tests/test_ops.py -v
```
*You should see all the tests pass with green checkmarks, showing that the security logic (401s and 403s) is functioning properly.*

## How to Test
To show the actual API rejecting and accepting requests, you can spin up the server and use postman to hit the api.

**1. Generate an API Key and Event**
First, use the CLI tool to create a new event and generate an access key for it:
```bash
uv run python -m app.cli admin create-client --event-name "Live Demo Event"
```
*(Copy the generated `API Key` from the output—you will need it in step 3!)*

**2. Start the Server**
Start the FastAPI server in your terminal:
```bash
uv run uvicorn app.main:app --reload
```

---

https://github.com/user-attachments/assets/f577d2fa-b976-4fac-b5ce-b17b4f3a741e

---

**3. Demonstrate the Security with `curl`**
Open a second terminal window.

**Demo A: Unauthenticated Request (Fails)**
Try to access the talks list without providing a key. This proves the endpoint is locked down.
```bash
http://localhost:8000/ops/talks
```
*Output should show a `401 Unauthorized` error.*

https://github.com/user-attachments/assets/ac735cf9-62d3-4b54-988a-9fb9289d244e

---

**Demo B: Authenticated Request (Succeeds)**
Now, add the headers key and value
```
key: X-API-Key
value: <API-KEY> you got in step 1
```
Try to access the talks list now.
```bash
http://localhost:8000/ops/talks
```
*Output should show a `200 OK` status and return an empty list `[]` (since we just created the event and haven't added any talks to it yet).*

https://github.com/user-attachments/assets/386970f1-d24e-4ba3-a3ce-776ef577950b

---

This clearly demonstrates both the new tests and the actual live security mechanisms working exactly as intended!

## Summary by Sourcery

Introduce authenticated operator APIs, database models, and CLI tooling for managing clients and events.

New Features:
- Expose an /ops API router providing read-only endpoints for listing talks, fetching talk details with jobs, and retrieving job details for authorized operator clients.
- Add a VEditor CLI entrypoint with an admin create-client command to create API clients scoped to events and print their API keys.
- Define SQLAlchemy models, Pydantic schemas, and Alembic migrations for events, clients, talks, jobs, and reviews, establishing relationships and status fields.
- Implement API-key based authentication and per-event authorization for operator clients using hashed keys and scoped event access.
- Introduce a finite state machine for talk status transitions to enforce valid workflow states.

Build:
- Register a veditor console_script entrypoint in the project configuration to expose the CLI.

Tests:
- Add unit and integration tests for ops endpoints covering authorized/unauthorized access and event scoping behaviour.
- Add model-level tests validating relationships, required fields, and client persistence with event scopes.
- Add CLI tests for create-client argument validation, event selection, and session lifecycle.
- Add authentication tests for API key hashing, client lookup, and event access checks.
- Add state transition tests verifying allowed and forbidden talk status changes.


Saksham-Sirohi (approved):
some smaller things can be fixed later

PR #139, opened 2026-08-17
feat: storage and local disk backend

### LINKED ISSUE #130: [Phase 2] `StorageBackend` interface + `LocalDiskBackend`
## Description
Every later phase (pipeline, retention, ingest, S3 migration) depends on one contract for touching files, plus a working v1 implementation of it. These are separated conceptually but shipped together here since the interface is only really validated once something implements it.

## Scope
- [ ] `app/storage.py`: define `StorageBackend` (ABC/Protocol) with `put`, `get`, `url`, `delete`, `exists`, `free_bytes`.
- [ ] Document semantics: overwrite behavior on `put`, missing-key behavior on `get`, `key` naming convention (`{talk_id}/{stage}/{filename}`).
- [ ] Implement `LocalDiskBackend(StorageBackend)` rooted at `DATA_DIR`, with `raw/`, `cut/`, `preview/`, `final/` per talk.
- [ ] `put()` creates missing directories; `get()` raises a clear error on missing keys; `delete()` is idempotent; `exists()` is a cheap check.
- [ ] `free_bytes()` returns real free space via `shutil.disk_usage(DATA_DIR)` — this is load-bearing for the Phase 5 disk guard, so cover it with a test against the actual filesystem, not a mock.

## Outcome
A stable, dependency-free storage contract exists, and it's backed by a real, runnable local-disk implementation. Every other phase — pipeline output, ingest, retention deletes, the eventual S3 swap — codes against this contract instead of touching the filesystem directly, which is what makes "swap backends without rewriting pipeline code" actually true rather than aspirational.

## Acceptance Criteria
- `app/storage.py` has zero FastAPI/Celery/DB imports.
- A file written via `put()` lands at `DATA_DIR/{talk_id}/{stage}/{filename}` and is byte-identical when retrieved via `get()`.
- `free_bytes()` matches `shutil.disk_usage`/`df` on the same directory in a test.
- Deleting a missing key does not raise.

ViRUS-0-0:
# Closes #130 

## Description
This PR implements **Phase 2: Storage & Ingest** by introducing the foundational storage contract and its first concrete implementation. 

To ensure our pipeline code remains decoupled from physical file operations (and to make migrating to S3/MinIO trivial in the future), all file interactions will now go through the `StorageBackend` interface.

## Changes
- **Configuration (`app/config.py`)**: Added `data_dir` and `storage_backend` settings.
- **Storage Interface (`app/storage.py`)**: 
  - Defined the `StorageBackend` Protocol with standard operations: `put`, `get`, `url`, `delete`, `exists`, and `free_bytes`.
  - Added `StorageKeyNotFoundError` to prevent native `FileNotFoundError` exceptions from leaking OS-level path structures.
- **Local Disk Implementation (`LocalDiskBackend`)**:
  - Implements the contract for local storage (v1).
  - **Atomic Writes**: `put()` writes to a hidden temporary file and uses `os.replace` to move it to the final key. If a massive video transcode crashes halfway through, the destination file won't be corrupted with a partial write.
  - Implements `free_bytes()` checking `shutil.disk_usage` to support the disk-space guards planned for Phase 5.
- **Testing (`tests/test_storage.py`)**: Added a comprehensive test suite (9 tests) covering `bytes` and `Path` ingest, idempotent deletions, correct exception handling, and a simulated write interruption to prove the atomic write guarantee.

## Testing Instructions
1. Run the new storage unit tests:
   ```bash
   uv run pytest tests/test_storage.py -v
   ```
2. Run the full suite to verify the `pipeline/` purity architectural guard remains unbroken:
   ```bash
   uv run pytest
   ```

## Summary by Sourcery

Introduce the storage abstraction and local disk backend needed to decouple pipeline file handling from the underlying storage system.

New Features:
- Add a storage backend contract and local disk implementation supporting file ingestion, retrieval, deletion, existence checks, URLs, and free-space reporting.

Enhancements:
- Provide atomic local storage writes, path traversal protection, and consistent missing-key errors.

Tests:
- Add unit coverage for local storage operations, error handling, disk-space reporting, path validation, and interrupted writes.


tests/test_storage.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Compare against direct shutil call
+    expected_free = shutil.disk_usage(tmp_path).free
+
+    # They should be very close, typically identical, but handle slight changes
+    # over the execution time if disk is active. We just ensure it's a valid integer
+    # roughly matching the disk usage.
+    assert isinstance(free, int)
+    assert free > 0
+    assert abs(free - expected_free) < 1024 * 1024 * 10  # within 10MB diff
```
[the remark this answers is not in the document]
ViRUS-0-0: Disk space on a non-isolated disk fluctuates constantly due to background processes, OS logs, etc. If any process writes or deletes a file in the time between calculating expected_free and free, the test would fail. The 10MB margin was intentionally added to handle this reality. Asserting exact equality would make this test highly flaky in most environments (especially CI).

tests/test_storage.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def test_free_bytes(storage_backend: LocalDiskBackend, tmp_path: Path):
+    # Call free_bytes
+    free = storage_backend.free_bytes()
+
+    # Compare against direct shutil call
+    expected_free = shutil.disk_usage(tmp_path).free
+
+    # They should be very close, typically identical, but handle slight changes
```
Rachit7168: No need for explaining steps and unnecessary comments

ViRUS-0-0: fixed

app/storage.py
```diff
@@ -0,0 +1,135 @@
...
+class StorageKeyNotFoundError(Exception):
+    """Raised when a requested key is not found in the storage backend."""
+
+    def __init__(self, key: str):
+        super().__init__(f"Storage key not found: {key}")
+        self.key = key
+
+
+class StorageBackend(Protocol):
+    def put(self, key: str, source: Path | bytes) -> None:
```
Saksham-Sirohi, line 22: lot of empty functions

ViRUS-0-0: Since this class inherits from `typing.Protocol`, it's just defining an interface. The actual implementations will live in the classes that fulfil this protocol.

app/storage.py
```diff
@@ -0,0 +1,135 @@
+import os
+import shutil
+import tempfile
+from pathlib import Path
+from typing import Protocol
+
+
+class StorageKeyNotFoundError(Exception):
```
hongquan: It should be subclass of [`FileNotFoundError`](https://docs.python.org/3/library/exceptions.html#FileNotFoundError).

app/storage.py
hongquan: These functions are fine to be just stub, because the `StorageBackend` class will not be used directly.

app/storage.py
```diff
@@ -0,0 +1,135 @@
...
+                    tmp.write(source)
+                else:
+                    with open(source, "rb") as f_in:
+                        shutil.copyfileobj(f_in, tmp)
+            except Exception:
+                tmp.close()
+                os.unlink(tmp.name)
+                raise
+
+        # Atomically replace the target file
```
hongquan: I don't understand what do you mean here?
The `os.replace` is already atomic.

app/storage.py
```diff
@@ -0,0 +1,135 @@
...
+        try:
+            os.replace(tmp.name, target_path)
+        except Exception:
+            os.unlink(tmp.name)
```
hongquan: What if the exception above is due to the `tmp.name`? If you `unlink` it here, the `unlink` also raise exception.

app/config.py
```diff
@@ -10,6 +10,9 @@ class Settings(BaseSettings):
 
     redis_url: str = "redis://localhost:6379/0"
 
+    data_dir: str = "data"
+    storage_backend: str = "local"
```
hongquan: Is there any requirement for `storage_backend`, or any string is allowed?

hongquan (changes requested):

app/storage.py
ViRUS-0-0: I've updated it to inherit directly from FileNotFoundError instead of Exception.

app/storage.py
ViRUS-0-0: `os.replace` is inherently atomic so the comment was redundant. I've removed it.

app/storage.py
ViRUS-0-0: It would throw a FileNotFoundError and mask the original error. I've updated both cleanup blocks to use `Path(tmp.name).unlink(missing_ok=True)` so it safely ignores missing files.

app/config.py
ViRUS-0-0: Good point, it shouldn't accept just any arbitrary string here. I've updated the type hint to use `Literal["local"]` so it strictly enforces the allowed backend options.

Saksham-Sirohi (changes requested):
Key convention is not documented. Issue 130 asks for {talk_id}/{stage}/{filename} in the contract. Protocol docstrings cover overwrite/missing-key, not the key scheme.

Directory keys leak through the file contract. After put("talk_1/raw/video.mp4"):

exists("talk_1") is True
get("talk_1") returns a directory Path instead of StorageKeyNotFoundError
delete("talk_1") now rmtrees the whole tree
That last part is useful for retention, but exists/get treating prefixes as objects will not match a future S3 backend. Either treat only files as keys, or document prefix-delete explicitly and make get/exists file-only.

storage_backend is unused. Literal["local"] is fine for now; nothing constructs a backend from settings. Acceptable for this phase if you are OK shipping config with no factory.

Saksham-Sirohi (approved):

PR #146, opened 2026-08-19
feat: path-based ingest: contract, validation & security

### LINKED ISSUE #136: [Phase 2] Path-based ingest: contract, validation & security
## Description
Per the roadmap, ingest assumes the client (or ops) has already placed the recording somewhere VEditor can see — a shared mount or pre-staged `DATA_DIR` — rather than uploading bytes over HTTP (multipart upload is explicitly Phase 7). Because the input is a client-supplied filesystem path, this is inherently adversarial input, not just a data-shape problem. Contract and security are one issue here because a path-based ingest endpoint that isn't validated against traversal and symlink escape isn't actually a finished feature — it's a vulnerability with a happy path attached.

Full API route + DB/RQ wiring happens in Phase 4; this issue's job is to produce a fully tested, callable service function that Phase 4 can drop straight into a route handler without re-solving path handling or security later.

## Technical Design

****Schema (`schemas.py`):****
```python
class RecordingIngestRequest(BaseModel):
    source_path: str | None = None
    relative_key: str | None = None

    @model_validator(mode="after")
    def exactly_one_path(self):
        if bool(self.source_path) == bool(self.relative_key):
            raise ValueError("exactly one of source_path or relative_key is required")
        return self
```

**Config:** `INGEST_ROOTS: list[Path]` — configured allowed base directories. `relative_key` resolves against these; `source_path` (absolute) must still land inside one of them after resolution.

**Service function** `app/ingest.py::stage_recording(talk_id: int, payload: RecordingIngestRequest, backend: StorageBackend) -> str`:

1\. Resolve the candidate path to an absolute, canonical path:
- `os.path.realpath()` (which resolves symlinks) — critical, since a naive `os.path.abspath()` does **not** resolve symlinks and would let a symlink inside an allowed root point outside `DATA_DIR` and pass a naive check.
- Normalize before comparison: collapse `..`, redundant slashes, and reject any path containing null bytes or unusual encodings up front.
2\. Check the resolved real path is a strict descendant of one of `INGEST_ROOTS` (string-prefix check on resolved paths is insufficient on its own — use `Path.is_relative_to()` or equivalent, and additionally verify the resolved parent directory actually exists and is a directory, not a file, to avoid edge cases with prefix-matching sibling names like `/data/ingest-evil` matching a naive prefix check against `/data/ingest`).
3\. If valid: copy the file into storage via `backend.put(key=f"{talk_id}/raw/{filename}", source=resolved_path)`.
4\. If invalid at any check: raise a specific `IngestPathRejectedError` — the route layer (Phase 4) maps this to `400`, with a message that does **not** echo back the internal resolved path.

## Scope
- [ ] `RecordingIngestRequest` schema with exactly-one-of validation.
- [ ] `INGEST_ROOTS` config.
- [ ] `stage_recording()` service function implementing the resolve → validate → stage flow above.
- [ ] `IngestPathRejectedError` custom exception.
- [ ] Happy-path tests: valid nested path under an allowed root stages correctly; both `source_path` and `relative_key` forms work.
- [ ] Adversarial tests (each as an explicit, named test case):
  - `../../etc/passwd`-style relative traversal
  - Absolute path entirely outside any `INGEST_ROOTS` entry
  - Symlink located inside an allowed root, target resolving outside `DATA_DIR`
  - Chained symlinks (symlink → symlink → outside target)
  - Trailing slash / double-slash path variants
  - URL-encoded or double-encoded traversal sequences (`%2e%2e%2f`)
  - Null-byte injection in the path string
  - Prefix-collision sibling directory (`/data/ingest-evil` vs. allowed root `/data/ingest`)
- [ ] Test asserting rejection happens **before** `backend.put()` is ever invoked — use a mock/spy on the backend, not just an HTTP-response-level check, so this stays a real guarantee even as the route layer around it changes in Phase 4.

## Out of Scope (this issue)
- The actual FastAPI route/endpoint and its RQ `detect` job trigger — that's Phase 4's `POST /talks/{id}/recordings`.
- Multipart/HTTP upload (Phase 7).
- Rate limiting on the endpoint (Phase 7).

## Outcome
There's a tested, callable, security-hardened ingest function that reliably turns "a path the client claims the file is at" into "a file safely inside VEditor's storage under the correct talk" — and cannot be abused as a path-traversal or arbitrary-file-read primitive by a malicious or misconfigured client, including via symlink tricks. Phase 4 can wire this directly into a route without re-deriving path-safety logic under the time pressure of also building the RQ/DB integration.

## Acceptance Criteria
- A valid `source_path`/`relative_key` under an allowed root stages the file into `talk/{id}/raw/`, byte-identical to the source, retrievable via `storage.get()`.
- Every adversarial test case listed above is rejected, and rejection is proven to occur before `backend.put()` is called.
- Providing neither or both of `source_path`/`relative_key` is rejected by schema validation with a clear error.
- Rejection error messages never contain the resolved internal filesystem path.

ViRUS-0-0:
# Closes #136 

## Description
This PR implements the Phase 2 path-based ingest validation, security contract, and in-process media verification. Because the ingest system relies on client-supplied filesystem paths (for recordings staged in shared mounts), all inputs are treated as untrusted and validated against strict boundaries before any staging or storage operations occur.

This PR introduces a secure `stage_recording` service function that safely canonicalizes and validates incoming paths, verifies container/stream integrity using PyAV, and prevents directory traversal, symlink escapes, and storage key collisions. This function is completely decoupled from the HTTP layer so Phase 4 can wire it directly into the `POST /talks/{id}/recordings` route.

## Key Changes
- **Configuration & Startup Validation**: Added `INGEST_ROOTS` to `Settings` with a Pydantic `field_validator` that rejects relative paths at startup and normalizes all allowed roots to canonical, absolute paths.
- **API Contract**: Introduced `RecordingIngestRequest` in `app/schemas.py` with an `@model_validator` enforcing that exactly one of `source_path` or `relative_key` is provided.
- **Secure Ingestion & Error Mapping (`app/ingest.py`)**: 
  - Resolves paths securely via `Path.resolve(strict=True)` and maps all OS-level errors (`OSError`, `RuntimeError`, permission errors, symlink cycles) to `IngestPathRejectedError` to prevent unhandled 500s.
  - Enforces strict root containment via `Path.is_relative_to(root)` and validates that `relative_key` is strictly relative.
  - Preserves root-relative paths in storage keys (`key = f"{talk_id}/raw/{rel_path}"`) to prevent silent key collisions for files with matching basenames across nested subdirectories.
- **In-Process Media Validation (PyAV)**:
  - Added `validate_media_file()` using PyAV (`av.open()`) to inspect media container headers in-process (< 2ms) and verify the presence of at least one video track before staging.
  - Pinned `av>=18.1.0` in `pyproject.toml` dependencies.

## Security Considerations
This PR explicitly defends against the following attack vectors:
- **Directory Traversal**: Relative path traversal attempts (e.g. `../` and URL-encoded traversal).
- **Symlink Escapes**: Malicious symlinks and chained symlinks resolving outside allowed roots.
- **Absolute Key Injection**: Absolute paths passed to `relative_key` attempting to bypass root prefixes.
- **Prefix Collisions**: Matching prefix spoofing (e.g. `/data/ingest-evil` vs `/data/ingest`).
- **Null-Byte Injections**: C-level string termination truncation via `\0`.
- **Non-Media / Audio-Only Uploads**: Rejecting text files, PDFs, binaries, and audio-only recordings before staging.
- **Side-Effect Invariants**: Asserts that `StorageBackend.put()` is never called when validation fails.

## Testing Done
- [x] Schema validation tests (`exactly_one_of` constraints).
- [x] Happy-path tests for both `source_path` and `relative_key`.
- [x] Storage key disambiguation test for identical basenames in different subdirectories.
- [x] Setting validator tests for relative vs. absolute `ingest_roots`.
- [x] In-process PyAV media rejection tests for non-media text files and audio-only files.
- [x] Comprehensive adversarial test suite asserting `mock_backend.put.assert_not_called()` across all attack vectors.

## Manual Test
Run `pytest` targeting the ingest test suite:
```bash
pytest tests/test_ingest.py -v
```

OUTPUT
```
tests/test_ingest.py::test_schema_validation_exactly_one PASSED          [  5%]
tests/test_ingest.py::test_happy_path_source_path PASSED                 [ 10%]
tests/test_ingest.py::test_happy_path_relative_key PASSED                [ 15%]
tests/test_ingest.py::test_disambiguated_storage_keys_same_basename PASSED [ 21%]
tests/test_ingest.py::test_adversarial_relative_traversal PASSED         [ 26%]
tests/test_ingest.py::test_adversarial_absolute_outside PASSED           [ 31%]
tests/test_ingest.py::test_adversarial_symlink_to_outside PASSED         [ 36%]
tests/test_ingest.py::test_adversarial_chained_symlinks PASSED           [ 42%]
tests/test_ingest.py::test_adversarial_trailing_slash_double_slash PASSED [ 47%]
tests/test_ingest.py::test_adversarial_url_encoded_traversal PASSED      [ 52%]
tests/test_ingest.py::test_adversarial_null_byte_injection PASSED        [ 57%]
tests/test_ingest.py::test_adversarial_prefix_collision PASSED           [ 63%]
tests/test_ingest.py::test_source_path_rejects_relative_path PASSED      [ 68%]
tests/test_ingest.py::test_relative_key_rejects_absolute_path PASSED     [ 73%]
tests/test_ingest.py::test_empty_ingest_roots_rejects_all PASSED         [ 78%]
tests/test_ingest.py::test_ingest_directory_rejected PASSED              [ 84%]
tests/test_ingest.py::test_settings_ingest_roots_validation PASSED       [ 89%]
tests/test_ingest.py::test_non_media_file_rejected PASSED                [ 94%]
tests/test_ingest.py::test_audio_only_file_rejected PASSED               [100%]
```

## Summary by Sourcery

Implement secure path-based recording ingestion with strict request validation, filesystem containment, video validation, and safe storage staging.

New Features:
- Add a path-based recording ingest contract supporting either an absolute source path or a root-relative key.
- Validate ingested media files and stage valid video recordings in storage using talk-scoped keys.

Bug Fixes:
- Reject directory traversal, symlink escapes, prefix-collision paths, null-byte inputs, invalid path forms, missing files, directories, and non-video media before storage.

Enhancements:
- Configure and validate absolute ingestion roots for secure path containment.
- Expose a dedicated ingestion path rejection error without leaking filesystem details.

Build:
- Add PyAV as a project dependency for media validation.

Tests:
- Add comprehensive schema, configuration, valid-ingest, media-validation, and adversarial path security tests.

## Summary by Sourcery

Implement secure path-based recording ingestion with strict validation, filesystem containment, video verification, and collision-safe staging.

New Features:
- Add a path-based recording ingest contract supporting either an absolute source path or a root-relative key.
- Validate video media and stage accepted recordings under talk-scoped storage keys.

Bug Fixes:
- Reject unsafe, invalid, missing, non-media, and audio-only paths before storage operations.

Enhancements:
- Configure canonical absolute ingest roots and expose a dedicated path-rejection error for invalid ingest requests.

Build:
- Add PyAV as a dependency for in-process media validation.

Tests:
- Add comprehensive schema, configuration, successful-ingest, media-validation, and adversarial path-security coverage.

### DIFF
--- a/app/ingest.py
+++ b/app/ingest.py
@@ -1,0 +1,47 @@
+from pathlib import Path
+
+from app.config import settings
+from app.schemas import RecordingIngestRequest
+from app.storage import StorageBackend
+
+
+class IngestPathRejectedError(ValueError):
+    pass
+
+
+def stage_recording(
+    talk_id: int, payload: RecordingIngestRequest, backend: StorageBackend
+) -> str:
+    path_str = payload.source_path or payload.relative_key
+    if "\0" in path_str:
+        raise IngestPathRejectedError("Invalid path")
+
+    target_path = Path(path_str)
+    resolved_path = None
+
+    # Resolve against roots
+    roots = [Path(r).resolve() for r in settings.ingest_roots]
+
+    if payload.source_path:
+        try:
+            candidate = target_path.resolve(strict=True)
+            if any(candidate.is_relative_to(root) for root in roots):
+                resolved_path = candidate
+        except FileNotFoundError:
+            pass
+    else:
+        for root in roots:
+            try:
+                candidate = (root / target_path).resolve(strict=True)
+                if candidate.is_relative_to(root):
+                    resolved_path = candidate
+                    break
+            except FileNotFoundError:
+                pass
+
+    if not resolved_path or not resolved_path.is_file():
+        raise IngestPathRejectedError("Invalid or missing ingest path")
+
+    key = f"{talk_id}/raw/{resolved_path.name}"
+    backend.put(key=key, source=resolved_path)
+    return key
--- a/app/ingest.py
+++ b/app/ingest.py
@@ -14,12 +14,14 @@ class IngestPathRejectedError(ValueError):
 def validate_media_file(path: Path) -> None:
     try:
         with av.open(str(path)) as container:
-            if not any(s.type == "video" for s in container.streams):
-                raise IngestPathRejectedError(
-                    "File is not a valid video (no video stream found)"
-                )
-    except (av.FFmpegError, IngestPathRejectedError) as exc:
+            has_video = any(s.type == "video" for s in container.streams)
+    except av.FFmpegError as exc:
         raise IngestPathRejectedError(f"Invalid media file: {exc}") from exc
+
+    if not has_video:
+        raise IngestPathRejectedError(
+            "File is not a valid video (no video stream found)"
+        )
 
 
 def stage_recording(
--- a/app/ingest.py
+++ b/app/ingest.py
@@ -23,6 +23,8 @@ def stage_recording(
     roots = [Path(r).resolve() for r in settings.ingest_roots]
 
     if payload.source_path:
+        if not target_path.is_absolute():
+            raise IngestPathRejectedError("source_path must be absolute")
         try:
             candidate = target_path.resolve(strict=True)
             if any(candidate.is_relative_to(root) for root in roots):
--- a/app/ingest.py
+++ b/app/ingest.py
@@ -1,5 +1,7 @@
 from pathlib import Path
 
+import av
+
 from app.config import settings
 from app.schemas import RecordingIngestRequest
 from app.storage import StorageBackend
@@ -7,6 +9,17 @@ from app.storage import StorageBackend
 
 class IngestPathRejectedError(ValueError):
     pass
+
+
+def validate_media_file(path: Path) -> None:
+    try:
+        with av.open(str(path)) as container:
+            if not any(s.type == "video" for s in container.streams):
+                raise IngestPathRejectedError(
+                    "File is not a valid video (no video stream found)"
+                )
+    except (av.FFmpegError, IngestPathRejectedError) as exc:
+        raise IngestPathRejectedError(f"Invalid media file: {exc}") from exc
 
 
 def stage_recording(
@@ -22,28 +35,38 @@ def stage_recording(
     # Resolve against roots
     roots = [Path(r).resolve() for r in settings.ingest_roots]
 
+    matched_root = None
     if payload.source_path:
         if not target_path.is_absolute():
             raise IngestPathRejectedError("source_path must be absolute")
         try:
             candidate = target_path.resolve(strict=True)
-            if any(candidate.is_relative_to(root) for root in roots):
-                resolved_path = candidate
-        except FileNotFoundError:
+            for root in roots:
+                if candidate.is_relative_to(root):
+                    resolved_path = candidate
+                    matched_root = root
+                    break
+        except OSError, RuntimeError:
             pass
     else:
+        if target_path.is_absolute():
+            raise IngestPathRejectedError("relative_key must be relative")
         for root in roots:
             try:
                 candidate = (root / target_path).resolve(strict=True)
                 if candidate.is_relative_to(root):
                     resolved_path = candidate
+                    matched_root = root
                     break
-            except FileNotFoundError:
+            except OSError, RuntimeError:
                 pass
 
-    if not resolved_path or not resolved_path.is_file():
+    if not resolved_path or not resolved_path.is_file() or not matched_root:
         raise IngestPathRejectedError("Invalid or missing ingest path")
 
-    key = f"{talk_id}/raw/{resolved_path.name}"
+    validate_media_file(resolved_path)
+
+    rel_path = resolved_path.relative_to(matched_root)
+    key = f"{talk_id}/raw/{rel_path}"
     backend.put(key=key, source=resolved_path)
     return key
--- a/.env.example
+++ b/.env.example
@@ -15,6 +15,10 @@ REDIS_URL=redis://localhost:6379/0
 DATA_DIR=data
 STORAGE_BACKEND=local
 
+# List of absolute directory paths that the app is allowed to ingest videos from.
+# Must be a valid JSON array of strings. Empty by default for security.
+INGEST_ROOTS='["/path/to/shared/mount", "/tmp/staging"]'
+
 # Optional Settings
 # SECRET_KEY=your_super_secret_key
 # DEBUG=True
--- a/app/config.py
+++ b/app/config.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from typing import Literal
 
 from pydantic_settings import BaseSettings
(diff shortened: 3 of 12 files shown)


SxxAq (approved):
LGTM Overall.

#### Minor Suggestions:
1. **Enforce Absolute vs Relative Path Semantics**:
   - When `payload.source_path` is provided, consider verifying `target_path.is_absolute()`. Currently, passing a relative string to `source_path` will resolve against the server process's CWD.
2. **Additional Edge Case Tests**:
   - Test that default/empty `settings.ingest_roots = []` safely rejects all ingestion attempts.
   - Test that attempting to ingest a directory (rather than a file) triggers `IngestPathRejectedError`.

also edit the pr summary as its not stacked anymore.

app/ingest.py
```diff
@@ -0,0 +1,47 @@
...
+
+    target_path = Path(path_str)
+    resolved_path = None
+
+    # Resolve against roots
+    roots = [Path(r).resolve() for r in settings.ingest_roots]
+
+    if payload.source_path:
+        try:
+            candidate = target_path.resolve(strict=True)
```
Saksham-Sirohi, line 45: `source_path` is an absolute client path. `Path(relative).resolve()` uses process CWD, so if CWD is inside an ingest root a relative `source_path` is accepted. Reject non-absolute `source_path` before resolve.

Saksham-Sirohi (changes requested):
One contract/security fix in `stage_recording`.

app/ingest.py
ViRUS-0-0: fixed

Saksham-Sirohi (changes requested):
ptal comment above

app/ingest.py
```diff
@@ -0,0 +1,72 @@
...
+
+def validate_media_file(path: Path) -> None:
+    try:
+        with av.open(str(path)) as container:
+            if not any(s.type == "video" for s in container.streams):
+                raise IngestPathRejectedError(
+                    "File is not a valid video (no video stream found)"
+                )
+    except (av.FFmpegError, IngestPathRejectedError) as exc:
+        raise IngestPathRejectedError(f"Invalid media file: {exc}") from exc
```
hongquan, line 19: What is the point of catching `IngestPathRejectedError` then raise the same type `IngestPathRejectedError`?

hongquan (changes requested):

app/ingest.py
ViRUS-0-0: Fixed! Separated the check so `try/except` only catches `av.FFmpegError` from container opening, and `IngestPathRejectedError` is raised directly when no video stream is found.

Saksham-Sirohi (approved):

Saksham-Sirohi (approved):

PR #151, opened 2026-08-21
feat: enforce storage boundary via AST architecture test

### LINKED ISSUE #138: [Phase 2] Enforce storage boundary (lint/test rule)
## Description
Architectural Invariant— "all file I/O goes through `storage.py`" — is a promise, not an enforced fact, unless something checks it mechanically. Without this, a future PR (very plausibly in Phase 3's pipeline work, where the temptation to just `open()` a file for a quick PyAV wrapper or manual byte-copy is highest) can quietly violate the boundary, and the violation typically isn't discovered until a retention delete misses a file because it was never routed through `storage.py`'s key convention.

## Technical Design
Two viable approaches — pick one, document the choice:

1\. **CI grep/AST test** (lower effort, good enough to start with): a pytest test (e.g. `tests/test_architecture.py`) that walks the `app/` tree excluding `storage.py`, parses each file's AST, and fails if it finds a call to `open()`, `shutil.copy*`, `shutil.move`, `shutil.rmtree`, `Path(...).write_bytes/write_text`, or `Path(...).read_bytes/read_text` **where the target path is plausibly under `DATA_DIR`** — in practice, simplest correct version is: fail on *any* direct use of these calls outside `storage.py`, and allow narrow, explicit exceptions (e.g. reading a config file, which isn't media) via an inline `# storage-boundary-exempt: reason` comment that the check looks for.

2\. **Custom `ruff`/`flake8` plugin rule** (higher effort, better DX — flags inline in the editor, not just in CI). Worth it only if the team is already comfortable maintaining a lint plugin.

Given Phase 2's scope, default to option 1 unless the team already has lint-plugin infrastructure.

Note: this check is orthogonal to the separate Phase 1 CI rule enforcing `pipeline/` import purity (no FastAPI/RQ/DB imports) — that rule checks *imports*, this one checks *filesystem calls*. Both matter but neither substitutes for the other; PyAV itself does raw file I/O internally when opening a container, which is fine — the rule targets `app/`-level code bypassing `storage.py`, not what PyAV does under the hood once handed a path by `storage.get()`.

## Scope
- [ ] Implement the AST-based (or grep-based, if AST is overkill) architecture test.
- [ ] Support an explicit, greppable exemption marker for legitimate non-media file I/O (config loading, log file writes, etc.) so the check doesn't produce false positives that get it disabled out of frustration.
- [ ] Wire into the Phase 1 CI job (same pipeline as lint/model tests/state-transition tests).
- [ ] Regression test: commit a deliberate violation to a scratch fixture module inside the test itself (not a real app file), assert the check fails on it, and confirm it does *not* fail on the exemption-marked case.
- [ ] Short note in code comments or a `docs/architecture.md` stub explaining the rule and why it exists (full `CONTRIBUTING.md` is Phase 8, but an early pointer avoids confused contributors before then).

## Out of Scope (this issue)
- General-purpose linting setup (assumed to exist from Phase 1).
- Enforcing the `pipeline/` import-purity rule — that's a separate Phase 1 CI check already covered there; this issue is storage-specific.

## Outcome
The storage-purity invariant becomes self-enforcing instead of relying on reviewer vigilance — a contributor who bypasses `storage.py`, including under the time pressure of later phases, is caught in CI before merge rather than discovered later during a retention bug where a file silently wasn't tracked by the key convention.

## Acceptance Criteria
- CI fails on a deliberately introduced direct-filesystem-I/O violation outside `storage.py`.
- CI passes on the current, clean codebase with zero false positives.
- The exemption marker mechanism exists and is tested (one legitimate exempted case does not fail the check).

SxxAq:
### Summary
Enforces the architectural invariant that all media and file I/O operations must strictly route through `app/storage.py`. This prevents accidental direct filesystem I/O in `app/` from bypassing storage backends and breaking retention/cleanup policies.

Fixes #138

### Key Changes
- **AST Architecture Test (`tests/test_storage_boundary.py`)**:
  - Scans all modules under `app/` (excluding `app/storage.py`) using Python's standard `ast` module.
  - Flags direct calls to builtin `open()`, `io.open()`, `shutil` file operations (`copy`, `move`, `rmtree`, etc.), `os` file operations (`remove`, `unlink`, `rename`, etc.), and `pathlib.Path` I/O methods (`read_bytes`, `write_text`, `unlink`, etc.).
  - Added unit test `test_app_storage_boundary` ensuring the current codebase conforms with zero violations.
  - Added regression test `test_storage_boundary_detects_violations` ensuring unexempted direct file I/O is caught with line numbers and details.
  - Added regression test `test_storage_boundary_respects_exemptions` confirming explicit exemptions pass.
- **Exemption Support**:
  - Allows legitimate non-media file I/O (e.g. config file loading, crash reporting) using the inline or preceding marker `# storage-boundary-exempt: <reason>`.
- **Documentation (`docs/architecture.md`)**:
  - Added architectural guidelines explaining the storage boundary invariant and how/when to use the exemption marker.

### Verification
Ran pytest against the test suite:
```bash
uv run pytest tests/test_storage_boundary.py tests/test_pipeline_imports.py tests/test_states.py tests/test_storage.py
============================== 17 passed in 0.15s ==============================

## Summary by Sourcery

Enforce routing of application media and persistent file I/O through the storage abstraction.

Enhancements:
- Enforce the storage boundary by adding an AST-based test that detects direct filesystem and file I/O calls throughout app/ while supporting documented, reasoned exemptions.

Documentation:
- Document the storage boundary invariant, its rationale, approved storage usage, and the exemption marker in the architecture guidelines.

Tests:
- Add coverage for detecting forbidden I/O, handling aliases, avoiding false positives, and honoring inline or preceding exemptions.

### DIFF
--- a/tests/test_storage_boundary.py
+++ b/tests/test_storage_boundary.py
@@ -1,0 +1,235 @@
+import ast
+from pathlib import Path
+
+EXEMPTION_MARKER = "# storage-boundary-exempt:"
+
+# Function names from modules or builtins that perform direct file I/O
+FORBIDDEN_BUILTINS = {"open"}
+FORBIDDEN_SHUTIL_FUNCS = {
+    "copy",
+    "copy2",
+    "copyfile",
+    "copyfileobj",
+    "copymode",
+    "copystat",
+    "copytree",
+    "move",
+    "rmtree",
+}
+FORBIDDEN_OS_FUNCS = {
+    "remove",
+    "unlink",
+    "rename",
+    "replace",
+    "rmdir",
+    "removedirs",
+    "mkdir",
+    "makedirs",
+}
+FORBIDDEN_PATH_METHODS = {
+    "read_bytes",
+    "read_text",
+    "write_bytes",
+    "write_text",
+    "unlink",
+    "rmdir",
+    "mkdir",
+    "open",
+}
+
+
+def is_exempt(lines: list[str], lineno: int) -> bool:
+    """Check if the line or the line immediately preceding it has an exemption marker."""
+    if 1 <= lineno <= len(lines):
+        # Check current line
+        if EXEMPTION_MARKER in lines[lineno - 1]:
+            return True
+        # Check preceding line if comment is placed immediately above
+        if lineno > 1:
+            prev_line = lines[lineno - 2].strip()
+            if prev_line.startswith(EXEMPTION_MARKER):
+                return True
+    return False
+
+
+def find_storage_violations(
+    source_code: str, file_path: str = "<unknown>"
+) -> list[str]:
+    """
+    Parse Python source code AST and find any unexempted direct filesystem/IO calls.
+    Returns a list of human-readable violation messages.
+    """
+    try:
+        tree = ast.parse(source_code)
+    except SyntaxError as e:
+        return [f"{file_path}:{e.lineno}: SyntaxError while parsing: {e}"]
+
+    lines = source_code.splitlines()
+    violations: list[str] = []
+
+    # Track imported names from forbidden modules (e.g., `from shutil import copyfile`)
+    imported_forbidden: dict[str, str] = {}
+
+    for node in ast.walk(tree):
+        if isinstance(node, ast.ImportFrom):
+            if node.module == "shutil":
+                for alias in node.names:
+                    if alias.name in FORBIDDEN_SHUTIL_FUNCS:
+                        imported_forbidden[alias.asname or alias.name] = (
+                            f"shutil.{alias.name}"
+                        )
+            elif node.module == "os":
+                for alias in node.names:
+                    if alias.name in FORBIDDEN_OS_FUNCS:
+                        imported_forbidden[alias.asname or alias.name] = (
+                            f"os.{alias.name}"
+                        )
+            elif node.module in ("io", "_io"):
+                for alias in node.names:
+                    if alias.name == "open":
+                        imported_forbidden[alias.asname or alias.name] = "io.open"
+
+    for node in ast.walk(tree):
+        if not isinstance(node, ast.Call):
+            continue
+
+        lineno = getattr(node, "lineno", 0)
+        if is_exempt(lines, lineno):
+            continue
+
+        func = node.func
+
+        # Case 1: Direct function calls like `open(...)` or imported `copyfile(...)`
+        if isinstance(func, ast.Name):
+            if func.id in FORBIDDEN_BUILTINS:
+                violations.append(
+                    f"{file_path}:{lineno}: Direct call to builtin '{func.id}()'. All file I/O must go through storage.py"
+                )
+            elif func.id in imported_forbidden:
+                orig_name = imported_forbidden[func.id]
+                violations.append(
+                    f"{file_path}:{lineno}: Direct call to '{orig_name}()' (via '{func.id}'). All file I/O must go through storage.py"
+                )
+
+        # Case 2: Attribute calls like `shutil.copy(...)`, `os.remove(...)`, `io.open(...)`, `Path(...).write_text(...)`
+        elif isinstance(func, ast.Attribute):
+            attr_name = func.attr
+
+            # Check `shutil.<func>` or `os.<func>` or `io.open`
+            if isinstance(func.value, ast.Name):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/test_storage_boundary.py [hunk 1]
```diff
@@ -0,0 +1,235 @@
...
+FORBIDDEN_PATH_METHODS = {
+    "read_bytes",
+    "read_text",
+    "write_bytes",
+    "write_text",
+    "unlink",
+    "rmdir",
+    "mkdir",
+    "open",
+}
```
ViRUS-0-0, line 42: `"open"` is included in `FORBIDDEN_PATH_METHODS`. Because the AST checker blindly flags *any* attribute method call matching these names (since it can't evaluate types dynamically), it will flag perfectly valid `.open()` calls on completely unrelated objects (e.g., `webbrowser.open()`, `PIL.Image.open()`, `zipfile.ZipFile().open()`).

**Resolution:** Remove `"open"` from `FORBIDDEN_PATH_METHODS`. The standalone `open()` builtin and `io.open()` are already caught in `FORBIDDEN_BUILTINS` and the module checks, which covers the vast majority of direct file I/O risks. The remaining `Path` methods (`write_text`, `read_bytes`, etc.) are unique enough to keep as heuristics.

tests/test_storage_boundary.py
```diff
@@ -0,0 +1,235 @@
...
+        return [f"{file_path}:{e.lineno}: SyntaxError while parsing: {e}"]
+
+    lines = source_code.splitlines()
+    violations: list[str] = []
+
+    # Track imported names from forbidden modules (e.g., `from shutil import copyfile`)
+    imported_forbidden: dict[str, str] = {}
+
+    for node in ast.walk(tree):
+        if isinstance(node, ast.ImportFrom):
```
ViRUS-0-0: The AST visitor checks `ast.ImportFrom` nodes to catch functions imported directly (e.g., `from shutil import copyfile`). However, it misses standard aliased module imports via `ast.Import`. If a developer uses `import shutil as sh` and then calls `sh.rmtree()`, the check fails because `func.value.id` will be `"sh"` instead of `"shutil"`.

**Resolution:** Add a handler for `ast.Import` to track module aliases, and resolve them when evaluating the attribute call further down.

tests/test_storage_boundary.py
```diff
@@ -0,0 +1,235 @@
...
+                )
+
+        # Case 2: Attribute calls like `shutil.copy(...)`, `os.remove(...)`, `io.open(...)`, `Path(...).write_text(...)`
+        elif isinstance(func, ast.Attribute):
+            attr_name = func.attr
+
+            # Check `shutil.<func>` or `os.<func>` or `io.open`
+            if isinstance(func.value, ast.Name):
+                module_name = func.value.id
+                if module_name == "shutil" and attr_name in FORBIDDEN_SHUTIL_FUNCS:
```
ViRUS-0-0: You also need to update the attribute check on line 102 to resolve the actual module name

ViRUS-0-0 (changes requested):
The PR LGTM, just minor changes required. Please take a look and resolve them.
Thanks

tests/test_storage_boundary.py
```diff
@@ -0,0 +1,276 @@
...
+}
+FORBIDDEN_PATH_METHODS = {
+    "read_bytes",
+    "read_text",
+    "write_bytes",
+    "write_text",
+    "unlink",
+    "rmdir",
+    "mkdir",
+}
```
Saksham-Sirohi, line 42: Leaving generic `.open()` out of this set is correct (`Image.open`, `webbrowser.open`). That still leaves a storage bypass: `Path(...).open()` / `p.open("wb")` and `os.open()`.

Please catch those without flagging unrelated `.open()` calls: add `open` to `FORBIDDEN_OS_FUNCS`, and flag `.open()` only when the receiver is a `Path(...)` / `pathlib.Path(...)` call (or a name bound to one).

Saksham-Sirohi (changes requested):
One remaining bypass in the architecture check.

Saksham-Sirohi (changes requested):
ptal comment above minor change required

Saksham-Sirohi (approved):

PR #156, opened 2026-08-21
feat: synthetic clip generator & PyAV test harness

### LINKED ISSUE #147: [Phase 3] Synthetic clip generator & PyAV test harness
## Description
Every module in this phase needs small media clips to test against, and a consistent way to inspect PyAV containers (duration, stream metadata, codec info) inside tests. The team decided against committing binary video fixtures to the repo — clips are instead **generated on-the-fly with PyAV** at test-collection or test-run time, used, and cleaned up automatically. This mirrors the common GStreamer pattern of a synthetic `videotestsrc` (`pattern=snow`, `pattern=pinwheel`, etc.), but implemented natively in PyAV so the test suite never needs GStreamer or its desktop-oriented dependency chain installed in CI.

Building the generator once, before any individual pipeline stage, means every later issue tests against the same synthesis helpers and the same assertion helpers instead of each one inventing its own throwaway clip-generation and comparison logic — which is exactly the kind of drift that makes six independently-written modules hard to trust as a set.

This also directly informs the open assumption already flagged in the roadmap: "PyAV covers all needed codecs/filters without shelling out to `ffmpeg`." The generator should be capable enough to exercise that risk (multiple codecs, audio+video, edge-case durations), not just produce one convenient short clip.

## Technical Design

**Generator module (`tests/pipeline/gen_video.py`):**

- `generate_clip(duration_s, *, has_video=True, has_audio=True, resolution=(320, 240), codec="h264", pattern="solid", fps=24, sample_rate=44100) -> Path` — synthesizes a short clip into a temp file and returns its path.
  - Video frames built via `av.VideoFrame.from_ndarray()` over numpy arrays. Support at least a couple of `pattern` options analogous to GStreamer's test patterns — e.g. `solid` (flat color), `gradient`, `noise` — since a completely static frame is a weak test signal for anything touching pixel data.
  - Audio built via `av.AudioFrame.from_ndarray()` — support a `tone` (fixed-frequency sine) and `silence` waveform, since `loudness.py`'s tests specifically need a known, measurable input level.
  - `has_video=False` / `has_audio=False` produce a container with only the other stream, for stream-presence edge cases.
- `generate_corrupt_clip() -> Path` — writes a valid-looking container header and then truncates/corrupts the body, for negative-path testing (`detect.py` and any module opening a container directly).
- `generate_mismatched_duration_clip(scheduled_start, scheduled_end, offset_s) -> Path` — thin wrapper around `generate_clip` that produces a clip whose duration deliberately falls outside a given scheduled window, paired with that window, so `detect.py`'s tests don't hand-roll the offset math themselves.
- All generated files are written under pytest's `tmp_path` (or an equivalent fixture-scoped temp dir) and are never written into the repo tree — nothing under `tests/pipeline/` is a binary asset.

**Investigate PyAV's own test-source support before hand-rolling everything:** confirm whether PyAV exposes anything built-in equivalent to GStreamer's `videotestsrc` patterns before committing to a fully manual numpy-frame approach — if it does, prefer it for simplicity; if not, the manual `from_ndarray` approach above is the fallback. Document the finding either way in the module docstring.

**Fallback plan (documented, not built by default):** if PyAV-native generation proves awkward for a specific case (e.g. an unusual codec PyAV can't easily encode a synthetic source into), the team may generate that one clip once with GStreamer locally (`gst-launch-1.0 videotestsrc ...`) and check in **the generation script**, not the resulting binary — CI still produces the actual test bytes on the fly. This keeps the "never upload/store real video, never commit binaries" rule intact even for the exception case.

**Helper module (`tests/pipeline/helpers.py`):**
- `open_and_inspect(path) -> ClipInfo` (duration, has_video, has_audio, codec names, resolution if video) — a thin wrapper so individual test files don't each hand-roll PyAV container introspection.
- `assert_duration_close(path_a, path_b, tolerance_seconds=...)` — used repeatedly across `cut`/`preview`/`transcode` tests where exact frame-accurate duration match isn't guaranteed (keyframe-boundary seeking, encoder rounding).
- `assert_playable(path)` — opens the container and decodes at least one frame from each stream, catching "file exists but is broken" failures that a bare `exists()` check would miss.

### Scope
- [ ] `tests/pipeline/gen_video.py` with `generate_clip`, `generate_corrupt_clip`, `generate_mismatched_duration_clip` as above.
- [ ] Spike/decision on whether PyAV has built-in test-pattern support vs. manual `from_ndarray` synthesis — documented in the module docstring either way.
- [ ] `tests/pipeline/helpers.py` with the three helper functions above.
- [ ] All generated clips written to a temp directory and cleaned up automatically after each test (no leftover files on disk, no binaries committed to the repo).
- [ ] A smoke test (`test_gen_video.py`) that generates one of every clip shape (standard, video-only, audio-only, corrupt, mismatched-duration) and asserts basic expected properties via the helpers (duration > 0 where applicable, correct stream presence, corrupt clip raises on open) — this is what actually proves the generator and helpers work together before six other issues start depending on them.
- [ ] README note (in `tests/pipeline/`) documenting the "no checked-in video fixtures, no uploaded video content" rule explicitly, so future contributors don't reintroduce binary fixtures out of habit.

## Out of Scope (this issue)
- Any pipeline-stage logic itself (detect/cut/preview/loudness/transcode/publish) — this issue only ships the generator and test infrastructure.
- Full-length (60-minute) sample files — those are for manual/beta testing (Phase 8) using a real (non-committed) sample, not the on-the-fly unit-test generator.
- Any GStreamer dependency in CI — GStreamer, if used at all, stays a one-time local/manual fallback per the plan above, never a CI requirement.

## Outcome
Every subsequent Phase 3 issue tests against freshly-synthesized, known-good (and known-bad) media generated at run time, using shared synthesis and assertion helpers instead of duplicated PyAV boilerplate — and instead of any binary video ever living in the repo or being uploaded anywhere. The generator's coverage (multiple codecs/patterns, audio+video combinations) also front-loads a real signal on whether PyAV's filter/encode path has gaps, before `loudness.py` or `transcode.py` are deep enough into implementation for a gap to be expensive to discover.

## Acceptance Criteria
- All generated clip shapes open successfully via `av.open()` and expose the expected streams (verified by the smoke test).
- The corrupt clip raises a catchable PyAV error on open, not a silent empty read.
- Helper functions are used by the smoke test, not left unexercised until later issues land.
- No video/audio binary files are committed to the repository; a test or CI check confirms `tests/pipeline/` contains no binary media assets.
- Generated temp files do not persist after the test session ends.

ViRUS-0-0:
## Closes #147 
This PR introduces the `gen_video.py` test helper, which uses PyAV and NumPy to synthesize deterministic, real-disk media files for testing the video processing pipeline. 

As established in our storage architecture discussions, while most of the application's test suite (ingest, database state, etc.) will use an in-memory `FakeStorageBackend` for speed, the Phase 3 pipeline tests require PyAV. Because PyAV delegates to FFmpeg C libraries, it strictly requires real, seekable file paths on a disk. This PR builds the standardized helpers to generate those real, temporary media files so the pipeline tests have safe, isolated, and consistent fixtures to consume.

We manually synthesize frames using NumPy rather than relying on PyAV's built-in `lavfi` virtual inputs to ensure full control over muxed audio/video clips, edge cases (audio-only or video-only), precise waveforms, and NumPy-readable pixel patterns.

## **Changes**
- Added `av` (PyAV) and `numpy` to `pyproject.toml` dependencies.
- Added `tests/pipeline/gen_video.py` containing the `generate_clip` helper.
- Supports configurable durations, frame rates, audio waveforms (`tone`, `silence`), and video patterns (`solid`, `gradient`, `noise`).
- Includes helpers for edge-case testing (e.g., `generate_corrupt_clip`, `generate_mismatched_duration_clip`).

## **Demo / How to Test**
First, verify that the automated tests pass successfully:
```bash
uv run pytest
```

Then, if you want to visually inspect the different output patterns, you can run this script to generate a set of demo fixtures locally:

```bash
uv run python - <<'PY'
from pathlib import Path
from tests.conftest import generate_clip

out = Path("/tmp/veditor-media-demo")
out.mkdir(exist_ok=True)

print(generate_clip(3, pattern="solid", output_dir=out))
print(generate_clip(3, pattern="gradient", output_dir=out))
print(generate_clip(3, pattern="noise", output_dir=out))
print(generate_clip(3, has_audio=False, pattern="gradient", output_dir=out))
print(generate_clip(3, has_video=False, audio_waveform="tone", output_dir=out))
PY

open /tmp/veditor-media-demo
```

## Video

https://github.com/user-attachments/assets/1c9c1027-4b13-41ac-9aec-ad2117348e29

## Summary by Sourcery

Provide deterministic runtime-generated media fixtures and validation utilities for pipeline testing.

New Features:
- Add configurable PyAV and NumPy helpers for generating deterministic audio/video test clips, including audio-only, video-only, corrupt, and duration-mismatched media.
- Add inspection and validation utilities for media duration, streams, codecs, resolution, and playability.

Enhancements:
- Standardize runtime-generated pipeline media fixtures in pytest temporary directories instead of storing binary assets in the repository.

Build:
- Add PyAV and NumPy project dependencies and update the lockfile.

Documentation:
- Document the policy of generating pipeline media fixtures at test runtime rather than committing binary files.

Tests:
- Add coverage for standard, audio-only, video-only, corrupt, and mismatched-duration generated clips, plus checks that no binary media assets are committed.

## Summary by Sourcery

Provide deterministic runtime-generated media fixtures and validation utilities for pipeline testing.

New Features:
- Add deterministic PyAV and NumPy helpers for generating configurable audio/video, audio-only, video-only, corrupt, and duration-mismatched media fixtures.
- Add media inspection, playback validation, and duration-comparison utilities for pipeline tests.

Enhancements:
- Standardize runtime-generated media fixtures in temporary directories instead of committing binary test assets.

Build:
- Add PyAV and NumPy dependencies and update the lockfile.

Tests:
- Add coverage for generated media variants, validation failures, duration mismatches, and the absence of committed media files.

## Summary by Sourcery

Provide deterministic runtime-generated media fixtures and validation utilities for pipeline testing.

New Features:
- Add deterministic PyAV and NumPy helpers for generating configurable audio/video, audio-only, video-only, corrupt, and duration-mismatched test media.
- Add media inspection, playback validation, and duration-comparison utilities for pipeline tests.

Enhancements:
- Standardize pipeline media fixtures as runtime-generated files rather than committed binary assets.

Build:
- Add PyAV and NumPy dependencies and update the lockfile.

Tests:
- Add coverage for generated media variants, validation failures, duration mismatches, and the absence of committed media files.


pyproject.toml
```diff
@@ -13,6 +13,7 @@ dependencies = [
     "alembic>=1.19.1",
     "psycopg[binary]>=3.3.4",
     "pydantic-settings>=2.15.0",
+    "av>=18.1.0",
```
Saksham-Sirohi, line 16: `av` is only imported from `tests/pipeline/`. Move it into the `dev` group next to `numpy` so production installs do not pull the FFmpeg stack. Add it to runtime deps when `app/` actually imports PyAV.

tests/test_gen_video.py
```diff
@@ -0,0 +1,150 @@
...
+    assert info.has_audio
+    assert info.duration is not None
+    assert info.duration > 0
+    assert_playable(path)
+
+
+def test_corrupt_clip_raises_on_open(tmp_path: Path):
+    path = generate_corrupt_clip(output_dir=tmp_path)
+
+    with pytest.raises(av.FFmpegError):
```
Saksham-Sirohi, line 96: `av.open()` often succeeds on a truncated-but-valid `ftyp` and only fails at decode, depending on the FFmpeg build. Assert that `assert_playable(path)` fails (not only that open raises), otherwise `generate_corrupt_clip` is not a stable fixture for later detect tests.

Saksham-Sirohi (changes requested):
Two file changes in this PR.

tests/test_gen_video.py
ViRUS-0-0: fixed

Saksham-Sirohi (changes requested):
ptal comment above: minor change required; looks good otherwise

tests/pipeline/test_gen_video.py
```diff
@@ -0,0 +1,150 @@
+from datetime import UTC, datetime, timedelta
+from pathlib import Path
+
+import av
+import pytest
+
+from tests.pipeline.gen_video import (
```
hongquan: The stuff in "tests" folder should be run pytest

pyproject.toml
ViRUS-0-0: @Saksham-Sirohi Added the pyAV dependency back from dev dependency as it is being used in ingest pr. After that is merged, I will resolve conflicts resulting in only changes in the test folder

Saksham-Sirohi (approved):

PR #157, opened 2026-08-24
feat(pipeline): detect.py pipeline script

### LINKED ISSUE #148: [Phase 3] `detect.py`: validate recording against talk schedule
> [!NOTE] 
> - **Depends on:** #147

## Description
This is the pipeline's first gate: given a raw recording and a talk's scheduled start/end, confirm the file plausibly belongs to that talk before any further compute is spent on it. Per the roadmap, this is validation only — pass/fail plus metadata — not batch auto-matching (explicitly deferred). It runs using PyAV container metadata (duration, stream presence) rather than an `ffprobe` subprocess call, which is the core Phase 3 constraint this whole phase exists to enforce.

## Technical Design

**Function signature:** `pipeline/detect.py::detect(file_path: Path, scheduled_start: datetime, scheduled_end: datetime) -> DetectResult`

**`DetectResult`** — a plain dataclass (no Pydantic, no DB model — `pipeline/` stays framework-free):
```python
@dataclass
class DetectResult:
    passed: bool
    actual_duration_seconds: float
    has_video: bool
    has_audio: bool
    reason: str | None  # populated when passed=False
```

**Validation logic:**
1. Open the container via `av.open(file_path)`. If this raises (corrupt/unreadable file), catch it and return `passed=False, reason="unreadable container"` — never let a PyAV exception escape this function uncaught, since callers (the eventual RQ task in Phase 4) shouldn't need to know PyAV's exception types.
2. Read duration from container metadata; compare against the scheduled window. Duration doesn't need to match exactly — the check is whether the recording's duration is consistent with the scheduled window (e.g. within a configurable tolerance, since recordings often start a little early/late relative to the schedule). Document the tolerance as a named constant, not a magic number inline.
3. Check for at least one video stream present (a talk recording with no video stream at all is very likely the wrong file). Audio-only is a hard fail here, not a warning — surfacing a wrong-file case clearly at this early gate is the entire point of `pending_approval`.
4. Return a populated `DetectResult` in all cases — this function should never raise for "the recording is wrong," only for truly exceptional cases (e.g. the file path itself doesn't exist, which is arguably a caller bug, not a detect-logic case — decide and document which of these two cases owns that check).

## Scope
- [ ] `pipeline/detect.py::detect()` implementing the above.
- [ ] `DetectResult` dataclass.
- [ ] Duration-tolerance constant, documented with reasoning (not just a bare number).
- [ ] Unit tests using phase 3 issue 1 on-the-fly generator:
  - A clip generated within the scheduled window (via `generate_clip`) → `passed=True`.
  - A clip generated via `generate_mismatched_duration_clip` → `passed=False`, reason mentions duration.
  - An audio-only clip (`generate_clip(has_video=False)`) against a schedule expecting video → `passed=False`, reason mentions missing video.
  - `generate_corrupt_clip()` → `passed=False`, reason mentions unreadable, and no exception escapes the test.
  - Nonexistent file path → documented, tested behavior (either a clean `DetectResult` failure or a deliberate raise — pick one and test it explicitly, don't leave it implicit).

## Out of Scope (this issue)
- Batch auto-matching / drop-folder detection (explicitly deferred in the roadmap).
- Anything DB- or API-related (the `pending_approval` state transition itself is Phase 4's job — this issue only produces the `DetectResult` that Phase 4's task will act on).

## Outcome
A pure, exception-safe validation function that reliably distinguishes "this file plausibly belongs to this talk" from "this is wrong or unreadable," with a result shape rich enough that Phase 4 can surface a clear reason to the client instead of a bare pass/fail.

## Acceptance Criteria
- `detect()` runs standalone on a file path with no FastAPI/RQ/DB imports.
- A recording whose duration falls outside the scheduled window (beyond tolerance) is correctly flagged with a duration-related reason.
- A corrupt or unreadable file returns a failed `DetectResult`, never an unhandled exception.
- All cases listed in Scope are covered by named, individually-readable test functions in `tests/pipeline/test_detect.py`, each generating its own input clip via Issue #147 's generator rather than reading a checked-in file.

ViRUS-0-0:
# Closes #148 

> [!NOTE]
> - This is a stacked pr and will be ready to review after #156 merges

This module acts as the first automated validation gate after recording ingest. It inspects media files using PyAV, extracts stream metadata, and verifies that the container's duration aligns with the talk's scheduled window within a 5-minute tolerance.

## Key Changes

- **Pure Pipeline Function**: Implemented `detect()` in `app/pipeline/detect.py` with zero dependencies on FastAPI, RQ, SQLAlchemy, or the database.
- **PyAV Extraction**: Uses PyAV to extract container duration and verify the presence of video/audio streams without spawning subprocesses (`ffmpeg`/`ffprobe`).
- **Structured Return Type**: Returns a frozen `DetectResult` dataclass containing `passed`, `actual_duration_seconds`, stream flags (`has_video`, `has_audio`), and a failure `reason`.
- **Edge Case & Error Handling**:
  - Missing file / zero-byte file checks short-circuit cleanly before invoking PyAV.
  - Catches `FFmpegError`, `OSError`, and `ValueError` to handle corrupt or unreadable containers.
  - Distinguishes between missing duration metadata (`duration <= 0`) and actual duration mismatches.
  - Treats audio-only recordings as a validation failure (missing video stream) while allowing video-only clips (e.g. mic failures) to pass.
  - Accommodates real-world AV setup delays with a practical 300s (5m) duration tolerance instead of eagerly routing standard drift to manual review.

### Automated Unit Tests
Run the pipeline test suite:
```bash
uv run pytest tests/pipeline/test_detect.py`
```

### Manual Verification

```bash
uv run python -c '
from datetime import datetime, timedelta
from pathlib import Path
from app.pipeline.detect import detect

start = datetime.now()
print(detect(
    file_path=Path("path/to/your_video.mp4"), #<- path to your video file
    scheduled_start=start,
    scheduled_end=start + timedelta(minutes=30) #<- can be changed to seconds as well
))
'
'
```

It should return a json output, something like this:

```python
DetectResult(passed=True, actual_duration_seconds=80.805442, has_video=True, has_audio=True, reason=None)
```

## Summary by Sourcery

Add automated recording validation for media integrity, stream presence, and scheduled-duration alignment.

New Features:
- Add a pure media detection pipeline that validates recording files against scheduled talk durations and stream requirements.
- Use PyAV to inspect media containers and return structured pass/fail results with duration and stream information.

Bug Fixes:
- Handle missing, empty, corrupt, unreadable, durationless, and audio-only recordings as explicit validation failures.

Enhancements:
- Allow video-only recordings to pass duration validation for downstream review.

Build:
- Add PyAV and NumPy development dependencies.

Documentation:
- Document the policy of generating temporary media fixtures instead of committing binary test assets.

Tests:
- Add synthetic media generation, inspection helpers, and coverage for valid, mismatched, audio-only, video-only, corrupt, missing, and empty recordings.

## Summary by Sourcery

Add automated validation for recording integrity, stream requirements, and alignment with scheduled talk durations.

New Features:
- Add a pure media detection pipeline that validates recording duration and stream presence against a scheduled talk window.
- Return structured validation results with media duration, audio/video availability, and failure reasons.

Bug Fixes:
- Handle missing, empty, corrupt, unreadable, durationless, and audio-only recordings as explicit validation failures.

Enhancements:
- Allow video-only recordings to pass when duration and other validation criteria are satisfied.
- Centralize container duration extraction for reuse by the detection pipeline and test helpers.

Tests:
- Add coverage for valid, mismatched-duration, stream-specific, corrupt, missing, and empty media files, including configurable tolerance boundaries.


SxxAq (approved):

app/pipeline/detect.py
```diff
@@ -0,0 +1,114 @@
...
+            has_video = bool(container.streams.video)
+            has_audio = bool(container.streams.audio)
+    except FFmpegError, OSError, ValueError:
+        return _failed("unreadable container")
+
+    if actual_duration <= 0:
+        return DetectResult(
+            passed=False,
+            actual_duration_seconds=actual_duration,
+            has_video=has_video,
```
Saksham-Sirohi, line 56: Python 3 cannot parse `except FFmpegError, OSError, ValueError`. Use `except (FFmpegError, OSError, ValueError):`. As written this file will not import and `detect()` cannot run.

Saksham-Sirohi (changes requested):
This module is a SyntaxError on Python 3 and cannot be imported.

app/pipeline/detect.py
ViRUS-0-0: In Python 3.14+, unparenthesized exception lists in except clauses (except A, B, C:) are valid syntax. This project targets `Python >= 3.14`, precommit's ruff format automatically normalizes `except (A, B, C):` to `except A, B, C:`. 

Manually adding parentheses causes the CI ruff format check to fail.

Saksham-Sirohi (approved):

PR #160, opened 2026-08-25
feat(pipeline): implement cut.py with stream-copy and re-encode fallback

### LINKED ISSUE #149: [Phase 3] `cut.py`: trim to scheduled bounds
> [!NOTE] 
> - **Depends on:** #147

## Description
Trims a raw recording down to the talk's scheduled start/end. This is the first Phase 3 module doing real media manipulation (not just metadata reads), and it's where the stream-copy-vs-re-encode decision — new complexity introduced by moving off the `ffmpeg` CLI, which handled this transparently via `-c copy` — first has to be made explicit in code.

## Technical Design

**Function signature:** `pipeline/cut.py::cut(input_path: Path, output_path: Path, start_seconds: float, end_seconds: float) -> None`

**Approach:**
1. Open input container, seek to `start_seconds` using PyAV's container-level seek (`container.seek()`), which seeks to the nearest keyframe at or before the target — document that output start may therefore be up to one GOP-length earlier than requested, which is why duration comparisons in tests need a tolerance, not exact equality.
2. Demux packets from the seek point up to `end_seconds`, remuxing directly into the output container **without decode/re-encode** when input and output codec/container combination is compatible (stream-copy path) — this is the fast, quality-lossless path and should be the default attempt.
3. Fall back to full decode → re-encode only when stream-copy isn't possible (e.g. codec not supported by the target container, or a precise non-keyframe cut boundary is explicitly required) — document this fallback trigger condition clearly, since it's a meaningful performance difference (stream-copy is fast; re-encode of a 60-minute clip is not) that a future contributor needs to understand, not just observe.
4. Both streams (if present) are cut in sync — video and audio should not drift relative to each other in the output.

## Scope
- [ ] `pipeline/cut.py::cut()` implementing seek → demux → remux, with stream-copy as the default path and re-encode as documented fallback.
- [ ] A `CutStrategy` enum or similar (`stream_copy`, `re_encode`) surfaced in the return or logged, so Phase 4's job logs can show which path was taken per talk (useful for debugging why one cut was fast and another wasn't).
- [ ] Unit tests, all inputs generated via Issue phase 3 issue 1's `gen_video.generate_clip`:
  - Cutting a generated standard clip produces output duration matching the requested window within tolerance.
  - Video-only and audio-only generated clips both cut correctly (no assumption that both streams exist).
  - Output remains playable (`assert_playable` from phase 3 Issue 1) after cutting.
  - At least one test explicitly exercises and asserts the stream-copy path was taken (not just that output looks right) — e.g. by checking the logged/returned strategy.
  - At least one test forces the re-encode fallback (e.g. by generating a clip with an incompatible codec via `generate_clip(codec=...)`, or explicitly by requesting a non-keyframe-aligned boundary if that's the trigger) and confirms output is still correct.

## Out of Scope (this issue)
- Any quality/bitrate adjustment — that's `preview.py`/`transcode.py`. `cut.py` trims, it doesn't re-quality.
- Loudness normalization — separate module.

## Outcome
A trim function that's fast by default (stream-copy) and correct as a fallback (re-encode), with the strategy choice visible rather than hidden — so a slow cut in production is diagnosable from the job log instead of a mystery.

## Acceptance Criteria
- `cut()` runs standalone on file paths with no FastAPI/RQ/DB imports.
- Output duration matches the requested `start`/`end` window within the documented tolerance.
- Stream-copy path is exercised and explicitly verified by at least one test (not just inferred from timing).
- Re-encode fallback path is exercised and explicitly verified by at least one test.
- All tests in `tests/pipeline/test_cut.py` pass, and none read a checked-in binary fixture.

SxxAq:
### Summary
Implements `app/pipeline/cut.py` to trim raw video/audio recordings to scheduled start/end bounds using PyAV.

Fixes #149
Depends on #156 (synthetic clip generator test harness)

### Technical Design & Highlights
- **Stream-Copy Fast Path (Default)**:
  - Uses container-level seeking (`container.seek()`) to the nearest keyframe before `start_seconds`.
  - Replicates stream headers via `add_stream_from_template()` and remuxes packets without decoding.
  - Offsets packet PTS and DTS consistently by the base DTS offset (`p.pts - base_offset`, `p.dts - base_offset`) to preserve timestamp ordering (`PTS >= DTS`) while ensuring the output stream starts at zero.
- **Re-Encode Fallback (`_cut_reencode`)**:
  - Automatically triggered when stream-copy is incompatible with output container/codecs or when `force_reencode=True` is passed for frame-accurate boundary cuts.
  - Decodes frames, filters by timestamp, and re-encodes clean output.
- **Observability**:
  - Surfaces a `CutStrategy` enum (`stream_copy` vs `re_encode`) in the return value to allow background workers and job logs to diagnose performance.
- **CLI Entry Point**:
  - Added `if __name__ == "__main__"` runner allowing direct terminal execution:
    `uv run python -m app.pipeline.cut input.mp4 output.mp4 --start 5 --end 20 [--reencode]`
- **Architectural Purity**:
  - Pure function taking/returning file paths with zero imports from FastAPI, Celery/RQ, or SQLAlchemy.

### Verification
Added comprehensive unit tests in `tests/pipeline/test_cut.py` using synthetic test clips:
- `test_cut_duration_matches_window`: Output duration matches target window within GOP tolerance.
- `test_cut_video_only` & `test_cut_audio_only`: Single stream video/audio cuts.
- `test_cut_reencode_fallback`: Verifies fallback re-encoding path.
- `test_cut_invalid_inputs`: Error handling on invalid bounds and missing files.
- All outputs validated playable with `assert_playable()`.

```bash
uv run pytest tests/pipeline/test_cut.py
============================== 5 passed in 3.19s ===============================

## Summary by Sourcery

Implement a standalone PyAV-based media cutting pipeline with fast stream copying and reliable re-encoding fallback.

New Features:
- Add media trimming with a default lossless stream-copy path and frame-accurate re-encode fallback.
- Expose the selected cut strategy and provide a command-line entry point for direct trimming operations.

Bug Fixes:
- Validate input paths and trim bounds to prevent invalid cuts and accidental input-file truncation.

Enhancements:
- Support trimming video-only, audio-only, and combined recordings while preserving playable output across codec/container combinations.

Tests:
- Add synthetic media tests covering stream-copy cuts, re-encoding, stream variants, codec resolution, invalid inputs, and output playability.

### CODE REVIEW

**Review by ViRUS-0-0 — CHANGES_REQUESTED** [review]
The branch has conflicts as the commits that you cherry picked has been merged to main. Please resolve conflicts and merge to main carefully, retaining the same functionality

File: pyproject.toml [hunk 1]
```diff
@@ -24,6 +24,8 @@ dev = [
     "httpx>=0.28.1",
     "ruff>=0.16.3",
     "pre-commit>=4.6.2",
+    "numpy>=2.5.2",
+    "av>=18.1.0",
```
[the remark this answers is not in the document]
ViRUS-0-0: Is already done in main branch. Please reflect the same

app/pipeline/cut.py
```diff
@@ -0,0 +1,251 @@
...
+
+    args = parser.parse_args()
+    strat = cut(
+        args.input,
+        args.output,
+        start_seconds=args.start,
+        end_seconds=args.end,
+        force_reencode=args.reencode,
+    )
+    print(f"Cut completed successfully using {strat.value} -> {args.output}")
```
ViRUS-0-0: It is not part of the issue requirements or acceptance criteria. In VEditor's design, pipeline functions are called programmatically by RQ workers, not via CLI. It was included solely as a developer convenience for ad-hoc manual testing and can be retained or removed without affecting the pipeline contract.
The CLI commands will be part of the future work flow having the base command "veditor cut"/"veditor detect" with configurable argument flags and so on.

app/pipeline/cut.py
```diff
@@ -0,0 +1,251 @@
...
+                    else (1.0 / av.time_base)
+                )
+                packet_time_s = (
+                    float(packet.pts * time_base)
+                    if packet.pts is not None
+                    else float(packet.dts * time_base)
+                )
+
+                if packet_time_s > end_seconds:
+                    continue
```
ViRUS-0-0, line 164: When trimming a short segment (e.g. 10 minutes) from a multi-hour recording, `packet_time_s > end_seconds` only executes `continue`. The demux loop continues reading every packet until the end of the file. In `_cut_reencode`, it decodes every remaining frame in the multi-hour video before discarding it, causing severe CPU and I/O bottlenecks.

Resolution: Track when all active streams have exceeded `end_seconds` and `break` out of the demux loop immediately.

app/pipeline/cut.py
```diff
@@ -0,0 +1,251 @@
...
+
+                    if frame_time_s < start_seconds or frame_time_s > end_seconds:
+                        continue
+
+                    if isinstance(frame, av.VideoFrame) and out_video is not None:
+                        frame.pts = video_frame_count
+                        frame.time_base = (
+                            Fraction(1, int(fps)) if int(fps) > 0 else Fraction(1, 24)
+                        )
+                        video_frame_count += 1
```
ViRUS-0-0, line 307: `int(fps)` truncates fractional framerates (such as NTSC standard 29.97 fps `Fraction(30000, 1001)` to `29 `or 23.976 fps to 23). Setting `frame.time_base = Fraction(1, int(fps))` creates a timebase mismatch against `out_video` (which was created with `rate=fps`), causing noticeable audio/video sync drift over long recordings.

Resolution: Compute the exact inverse fraction using `Fraction(1, 1) / Fraction(fps)` or `1 / fps` to retain fractional precision.

tests/pipeline/test_cut.py
ViRUS-0-0, line 1: @Saksham-Sirohi A doubt related to tests for the pipeline. Should the tests for pipeline be in a specific folder `tests/pipeline` or current way is fine?

ViRUS-0-0 (changes requested):

app/pipeline/cut.py
ViRUS-0-0: You can reference the working and code design in #161 & #157 

@SxxAq

ViRUS-0-0 (changes requested):
A working demo video of the script using the python shell calling the function should be added in the pr description. 

Also, cli commands such as "veditor cut"/"veditor detect" will be added in the future phase. Adding `if __name__ == "__main__":` with `argparse` here is out of scope and introduces dead boilerplate.

ViRUS-0-0 (approved):

app/pipeline/cut.py
```diff
@@ -0,0 +1,241 @@
...
+
+            if video_streams:
+                in_v = video_streams[0]
+                codec_name = in_v.codec_context.name or "libx264"
+                fps = in_v.average_rate or in_v.guessed_rate or 24
+                out_video = out_container.add_stream(codec_name, rate=fps)
+                out_video.width = in_v.codec_context.width
+                out_video.height = in_v.codec_context.height
+                out_video.pix_fmt = in_v.codec_context.pix_fmt or "yuv420p"
+                video_time_base = (
```
Saksham-Sirohi, line 246: `in_v.codec_context.name` is the decoder (`h264`), not an encoder (`libx264`). Map decoder names to encoders (and the same for audio, e.g. `aac`) or the forced-reencode path raises instead of cutting.

Saksham-Sirohi (changes requested):
The re-encode fallback will fail on typical H.264 inputs because it uses the decoder codec name as the encoder.

app/pipeline/cut.py
SxxAq: mapped

app/pipeline/cut.py
```diff
@@ -0,0 +1,275 @@
...
+    if start_seconds < 0:
+        raise ValueError(f"start_seconds must be non-negative: {start_seconds}")
+
+    if end_seconds <= start_seconds:
+        raise ValueError(
+            f"end_seconds ({end_seconds}) must be greater than start_seconds ({start_seconds})"
+        )
+
+    # storage-boundary-exempt: creating parent directory for pipeline output
+    out_path.parent.mkdir(parents=True, exist_ok=True)
```
Saksham-Sirohi, line 103: Reject identical input and output paths before opening the destination with `mode="w"`. `cut(source, source, …)` truncates the recording before it is read.

Saksham-Sirohi (changes requested):
Encoder mapping looks fixed. One remaining data-loss case.

app/pipeline/cut.py
```diff
@@ -0,0 +1,275 @@
...
+            if video_streams:
+                in_v = video_streams[0]
+                encoder_name = _resolve_video_encoder(in_v.codec_context.name)
+                fps = in_v.average_rate or in_v.guessed_rate or 24
+                out_video = out_container.add_stream(encoder_name, rate=fps)
+                out_video.width = in_v.codec_context.width
+                out_video.height = in_v.codec_context.height
+                out_video.pix_fmt = in_v.codec_context.pix_fmt or "yuv420p"
+                video_time_base = (
+                    Fraction(1, 1) / Fraction(fps) if fps else Fraction(1, 24)
```
ViRUS-0-0, line 247: When stream-copy fails due to container/codec incompatibility (e.g., trimming an input file containing `vp8`, `vp9`, or `vorbis` into an `.mp4` destination), `_cut_reencode` resolves the encoder directly from the input stream codec (`"vp8"` -> `"libvpx"`). Passing `"libvpx"` to `out_container.add_stream("libvpx")` on an MP4 container raises `ValueError: 'mp4' format does not support 'libvpx' codec`, causing the re-encode fallback to crash instead of producing the trimmed clip.

Wrap `add_stream` in a fallback that checks if the requested encoder is accepted by the output container, falling back to the container format's default video/audio codec (or `"libx264"` / `"aac"` for MP4 containers) when a `ValueError` is raised.

ViRUS-0-0 (changes requested):

ViRUS-0-0 (approved):

ArnavBallinCode (approved):

PR #168, opened 2026-08-26
feat(pipeline): implement audio loudness normalization with PyAV loudnorm

### LINKED ISSUE #153: [Phase 3] `loudness.py`: audio normalization
> [!NOTE] 
> - **Depends on:** #147

## Description
Normalizes audio loudness (EBU R128 / bs1770-style) using PyAV's filter graph API, rather than shelling out to `ffmpeg`'s built-in `loudnorm` filter via subprocess. This is the module most likely to expose a real gap in PyAV's filter coverage, since `loudnorm` on the CLI is a mature, single-flag, two-pass-capable filter, and PyAV's filter graph API may not expose an equivalent as directly — this is exactly the risk the roadmap's open assumption flags, and this issue is where that assumption gets tested for real rather than staying theoretical.

## Technical Design

**Function signature:** `pipeline/loudness.py::normalize(input_path: Path, output_path: Path, target_lufs: float = -16.0) -> None`

**Investigation required before implementation is "done":**
1. Confirm whether PyAV's `av.filter.Graph` exposes a direct loudness-normalization filter (e.g. wrapping libav's own `loudnorm` filter, since PyAV filter graphs can reference the same underlying libavfilter filters `ffmpeg` uses — the filter itself likely exists in libavfilter regardless of whether it's exposed identically through PyAV's Python API).
2. If a single-pass approach isn't clean via the graph API, implement the standard two-pass approach: first pass measures integrated loudness/true peak/loudness range, second pass applies gain correction based on those measurements — document whichever approach is used and why, since a future contributor debugging a loudness issue needs to know which path is live.
3. Whichever approach is used, write the decision and any PyAV filter-API limitations discovered into a short note in the module docstring — this directly feeds back into confirming or invalidating the roadmap's open assumption about PyAV filter coverage.

## Scope
- [ ] Spike/investigation of PyAV's filter graph API against the target loudness-normalization behavior — document findings before finalizing implementation approach.
- [ ] `pipeline/loudness.py::normalize()` implementing whichever approach (direct filter or two-pass measure+apply) the investigation lands on.
- [ ] Unit tests, using Issue #147 's generator with a known-level `tone`/`silence` waveform (this is precisely why the generator supports those waveform types — a synthetic tone at a known dB level is a far cleaner test input than any recorded clip would be):
  - Normalized output's measured loudness is closer to `target_lufs` than the unnormalized generated input (measured via a separate measurement pass in the test, not just trusting the implementation).
  - Video stream (if present in the generated clip) is unaffected — same duration, same frame count/codec as input — since this is an audio-only operation.
  - Audio-only generated clip (`has_video=False`) also normalizes correctly.

## Out of Scope (this issue)
- Applying loudness normalization as part of the `preview.py` path (previews are for visual review, not final audio delivery) — this module is invoked in the path toward `transcode.py`/publish.

## Outcome
Either a confirmed, working PyAV-native loudness normalization path, or a documented two-pass fallback — either way, the roadmap's open assumption about PyAV filter coverage gets a concrete, tested answer instead of remaining a guess going into Phase 4.

## Acceptance Criteria
- `normalize()` runs standalone with no FastAPI/RQ/DB imports.
- Output audio loudness measurably moves toward the target LUFS relative to a generated known-level input (verified by an independent measurement in the test, not by trusting the normalize function's own internal calculation).
- Video stream, if present, is unaffected (same duration, same frame count) by the audio-only operation.
- The module docstring documents which approach (direct filter vs. two-pass) was used and why.
- All tests in `tests/pipeline/test_loudness.py` pass.

SxxAq:
### Summary
Implements `app/pipeline/loudness.py` to normalize audio loudness to EBU R128 / BS.1770 standards using PyAV's native filter graph API.

Fixes #153

### Technical Design & Highlights
- **PyAV Native Filter Graph**:
  - Investigated and confirmed that PyAV's `av.filter.Graph` exposes libavfilter's `loudnorm` filter directly.
  - Implements single-pass loudness normalization (`I={target_lufs}:TP=-1.5:LRA=11`) with `abuffer -> loudnorm -> abuffersink`.
- **Lossless Video Preservation**:
  - When video streams are present, video packets are demuxed and remuxed via stream-copy (`add_stream_from_template()`) with zero video re-encoding overhead.
- **Audio Stream Sync**:
  - Normalized audio frames are encoded through AAC with synchronized PTS and sample rate timebases.
- **Observability & Standalone CLI**:
  - Includes CLI entrypoint: `uv run python -m app.pipeline.loudness input.mp4 output.mp4 --target-lufs -16.0`
- **Purity**:
  - Fully isolated with zero FastAPI, Celery/RQ, or database dependencies.

### Verification
Added unit tests in `tests/test_loudness.py`:
- `test_normalize_moves_loudness_towards_target`: Validates output duration and playability.
- `test_normalize_preserves_video_stream`: Asserts video codec, resolution, and duration are preserved.
- `test_normalize_audio_only`: Validates audio-only clips.
- `test_normalize_no_audio_stream_raises`: Validates error handling for audio-less media.
- `test_normalize_invalid_arguments`: Input and parameter boundary validation.

```bash
uv run pytest tests/test_loudness.py
============================== 5 passed in 2.82s ===============================

## Summary by Sourcery

Implement configurable EBU R128 loudness normalization for media files while preserving existing video streams.

New Features:
- Add a standalone PyAV-based audio loudness normalization API targeting configurable LUFS levels.
- Preserve video streams through stream-copy remuxing while producing normalized audio output for video and audio-only media.
- Provide validation for input paths, output paths, loudness targets, and missing audio streams.

Tests:
- Add coverage for loudness normalization, media playability and duration, video preservation, audio-only inputs, and invalid arguments.

### CODE REVIEW

File: app/pipeline/loudness.py [hunk 1]
```diff
@@ -0,0 +1,160 @@
...
+    parser.add_argument(
+        "--target-lufs",
+        type=float,
+        default=DEFAULT_TARGET_LUFS,
+        help=f"Target integrated loudness in LUFS (default: {DEFAULT_TARGET_LUFS})",
+    )
+
+    args = parser.parse_args()
+    normalize(args.input, args.output, target_lufs=args.target_lufs)
+    print(f"Audio normalized successfully -> {args.output}")
```
ViRUS-0-0: Unneeded CLI entrypoint block (`if __name__ == "__main__":`).
Pipeline modules are pure internal libraries orchestrated by background workers and tested via unit tests. CLI commands are centralized in [`app/cli.py`](./app/cli.py). Adding `if __name__ == "__main__":` with `argparse` here is out of scope and introduces dead boilerplate.

Delete the `if __name__ == "__main__":` block entirely to keep the module minimal and compliant with YAGNI.

In VEditor's design, pipeline functions are called programmatically by RQ workers, not via CLI. It was included solely as a developer convenience for ad-hoc manual testing and can be retained or removed without affecting the pipeline contract.

app/pipeline/loudness.py
```diff
@@ -0,0 +1,160 @@
...
+            audio_sample_count = 0
+
+            for packet in in_container.demux(in_a):
+                for frame in packet.decode():
+                    graph.push(frame)
+                    while True:
+                        try:
+                            out_frame = graph.pull()
+                        except av.FFmpegError, EOFError:
+                            break
```
ViRUS-0-0: Missing parentheses in `except` clause tuple.
`except av.FFmpegError, EOFError:` without parentheses is unidiomatic in Python 3 / PEP 8 and resembles legacy Python 2 comma syntax. Catching multiple exceptions must be specified as a tuple `except (av.FFmpegError, EOFError):`.

**Resolution**: Wrap the exception types in parentheses `(av.FFmpegError, EOFError)`.

app/pipeline/loudness.py
```diff
@@ -0,0 +1,160 @@
...
+                        for enc_packet in out_audio.encode(out_frame):
+                            out_container.mux(enc_packet)
+
+            # Flush filter graph
+            graph.push(None)
+            while True:
+                try:
+                    out_frame = graph.pull()
+                except av.FFmpegError, EOFError:
+                    break
```
ViRUS-0-0, line 145: Same here as the above comment

app/pipeline/loudness.py
```diff
@@ -0,0 +1,160 @@
...
+
+            # Mux video packets if present
+            if out_video is not None:
+                for packet in in_container.demux(video_streams[0]):
+                    if packet.dts is None:
+                        continue
+                    packet.stream = out_video
+                    out_container.mux(packet)
+
+            # Demux, filter, encode, and mux audio packets
```
ViRUS-0-0: Two-pass sequential demuxing with `in_container.seek(0)`.
Currently, all video packets are demuxed and muxed first, then `in_container.seek(0)` is invoked, followed by audio demuxing. Seeking may miss boundary samples or fail on non-seekable streams, and writing all video packets before any audio packets leads to un-interleaved containers.

**Resolution**: Demux active streams in a single pass without calling `seek(0)`.

app/pipeline/loudness.py
```diff
@@ -0,0 +1,160 @@
...
+            # 1. Video stream copy (if present)
+            out_video = None
+            if video_streams:
+                out_video = out_container.add_stream_from_template(video_streams[0])
+
+            # 2. Configure normalized audio stream
+            out_audio = out_container.add_stream("aac", rate=sample_rate)
+            out_audio.layout = (
+                in_a.codec_context.layout.name if in_a.codec_context.layout else "mono"
+            )
```
ViRUS-0-0: Audio layout fallback when `layout` metadata is missing.
If `in_a.codec_context.layout` is `None` but `channels == 2`, defaulting unconditionally to `"mono"` downmixes stereo tracks.

**Resolution**: Check channel count or default sensibly when layout name is absent.

ViRUS-0-0 (changes requested):
A working demo video of the script using the python shell calling the function should be added in the pr description. 

Also, cli commands such as "veditor cut"/"veditor detect" will be added in the future phase. Adding `if __name__ == "__main__":` with `argparse` here is out of scope and introduces dead boilerplate.

app/pipeline/loudness.py
```diff
@@ -0,0 +1,150 @@
...
+                        while True:
+                            try:
+                                out_frame = graph.pull()
+                            except av.FFmpegError, EOFError:
+                                break
+
+                            out_frame.pts = audio_sample_count
+                            out_frame.time_base = Fraction(1, sample_rate)
+                            audio_sample_count += out_frame.samples
+                            for enc_packet in out_audio.encode(out_frame):
```
Saksham-Sirohi, line 136: `except av.FFmpegError, EOFError` is a SyntaxError in Python 3. Use `except (av.FFmpegError, EOFError):`. Same on the flush loop below.

app/pipeline/loudness.py
```diff
@@ -0,0 +1,150 @@
...
+            graph.push(None)
+            while True:
+                try:
+                    out_frame = graph.pull()
+                except av.FFmpegError, EOFError:
+                    break
+
+                out_frame.pts = audio_sample_count
+                out_frame.time_base = Fraction(1, sample_rate)
+                audio_sample_count += out_frame.samples
```
Saksham-Sirohi, line 149: Same SyntaxError: `except av.FFmpegError, EOFError` must be `except (av.FFmpegError, EOFError):`.

Saksham-Sirohi (changes requested):
The module is invalid Python 3, and the audio/video mux path still has the issues already flagged.

app/pipeline/loudness.py
```diff
@@ -0,0 +1,153 @@
...
+                                except (av.FFmpegError, EOFError) as _:
+                                    break
+
+                                out_frame.pts = audio_sample_count
+                                out_frame.time_base = Fraction(1, sample_rate)
+                                audio_sample_count += out_frame.samples
+                                for enc_packet in out_audio.encode(out_frame):
+                                    out_container.mux(enc_packet)
+                    except (av.FFmpegError, EOFError) as _:
+                        continue
```
Saksham-Sirohi: Keep the inner `graph.pull()` catch for filter EOF. This outer `except (av.FFmpegError, EOFError): continue` also wraps `out_audio.encode()` and `mux()`, so a failed encode is skipped and the file is still written. Let encode/mux errors raise.

Saksham-Sirohi (changes requested):
The `except` syntax and NaN check are fixed. The outer audio catch still hides real encode/mux failures.

ViRUS-0-0 (approved):
LGTM

ArnavBallinCode (approved):

PR #189, opened 2026-09-03
feat(pipeline): implement PyAV-based transcode progress tracking

### LINKED ISSUE #183: [Phase 5] PyAV-based progress tracking during encode/decode
### Description
Transcoding a 60+ minute session can run for a long time with no feedback. This issue instruments the PyAV encode/decode loop to compute and persist a completion percentage as it runs, rather than leaving clients polling a silent status enum.

### Technical design
- In `transcode.py` (and any other PyAV stage worth instrumenting — `cut.py`/`preview.py` if their runtime is also non-trivial, though `transcode.py` is the priority per the roadmap's framing):
  - Read total container duration up front via PyAV's container metadata (`container.duration`).
  - During the decode/encode loop, track the current packet/frame's `pts` (presentation timestamp), convert to the same time base as `container.duration`, and compute `progress_ratio = current_pts / total_duration`.
  - Throttle updates to roughly every 5–10 seconds of wall-clock time (not every frame — avoid DB write amplification).
  - Each update: open a short-lived DB session, write `Job.progress_pct` (and bump `updated_at`), close the session immediately — per architectural invariant #4, the session must **not** remain open across the PyAV processing itself, only for these brief periodic writes.
- Guard against `container.duration` being unavailable/`None` for some inputs — fall back gracefully (e.g. `progress_pct` stays `null`/unset rather than crashing the job).

### Scope
- Progress computation logic within `transcode.py`.
- Periodic, short-lived DB writes of `Job.progress_pct` / `updated_at` from within the RQ job.
- Graceful fallback when duration metadata is unavailable.

### Out of scope
- Surfacing this data via the API response (Issue 8).
- Progress tracking for `cut.py`/`preview.py` unless trivially reusable from the same helper — if their runtime turns out to matter, treat as a follow-up rather than blocking this issue.
- Sub-stage progress (e.g. separate audio/video track progress) — a single overall ratio is sufficient for v1.

### Outcome
Long-running transcode jobs update a live, DB-persisted progress percentage throughout their execution, without violating the no-open-session-during-processing invariant.

### Acceptance Criteria
- Running a real multi-minute fixture transcode shows `Job.progress_pct` increasing across at least 3 distinct polls during the run.
- DB session is opened only briefly for each progress write — verified (e.g. via test/log instrumentation) that no session spans the PyAV processing call itself.
- An input with unavailable duration metadata does not crash the job; `progress_pct` degrades gracefully (stays null or last-known value).

### Dependency
Depends on Phase 3's `transcode.py` (PyAV pipeline core) and Phase 4's job/queue wiring being in place. Independent of Issues 1–6.

SxxAq:
## Summary
Closes #183.

Instruments the PyAV transcoding decode/encode loop to calculate presentation timestamp completion ratios, throttle database writes, and persist `Job.progress_pct` using short-lived database sessions.

---

## Key Changes

* **Database & Models**:
  * Added `progress_pct: Mapped[float | None] = mapped_column(Float, nullable=True)` to `models.Job`.
  * Added Alembic migration `0003_add_job_progress_pct.py` to add `progress_pct` to the `jobs` table.
  * Updated `schemas.JobBase` to include `progress_pct: float | None = None`.
* **Pipeline Core (`app/pipeline/transcode.py`)**:
  * Computes `progress_ratio = elapsed_pts_s / in_duration_s` and passes updates via `on_progress(pct)`.
  * Gracefully handles containers where duration is `None` or unavailable without crashing.
* **Task Worker Orchestration (`app/tasks.py`)**:
  * Updated `job_transcode` to track progress with a default 5.0-second wall-clock throttle.
  * Updates `Job.progress_pct` using brief, short-lived `SessionLocal()` instances, ensuring no database session remains open across media processing.
  * Sets `Job.progress_pct = 100.0` upon transcode completion.
* **Testing**:
  * Added `test_transcode_graceful_fallback_when_duration_unavailable` in `tests/pipeline/test_transcode.py`.
  * Added `test_transcode_progress_callback_updates_job_progress` in `tests/test_tasks.py`.

---

## Verification

```bash
# Run transcode, task, model, and boundary tests
$ uv run pytest tests/pipeline/test_transcode.py tests/test_tasks.py tests/test_models.py tests/test_storage_boundary.py
======================= 36 passed in 12.07s ========================

# Linting and formatting
$ uv run ruff check app/ tests/ migrations/versions/0003_add_job_progress_pct.py
$ uv run ruff format app/ tests/ migrations/versions/0003_add_job_progress_pct.py --check
All checks passed!

## Summary by Sourcery

Add durable, throttled progress tracking to PyAV transcoding jobs.

New Features:
- Persist transcode completion progress on jobs and expose it through the job schema.
- Report PyAV transcode progress during processing and mark completed jobs at 100%.
- Handle media with unavailable duration without failing transcoding.

Enhancements:
- Throttle progress persistence using short-lived database sessions so media processing does not retain an open session.

Tests:
- Add coverage for progress callback updates, completion status, monotonic progress reporting, and missing-duration fallback.

Chores:
- Add the database migration for the new job progress field.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added progress tracking for jobs, including percentage updates during processing.
  * Added timing information for talks, including source duration and cut boundaries.

* **Bug Fixes**
  * Improved progress reporting so updates are timely, steadily increasing, and completed accurately.
  * Prevented canceled or terminal jobs from receiving further updates or being incorrectly marked as failed.
  * Improved handling when source duration information is unavailable.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

File: app/tasks.py [hunk 1]
```diff
@@ -362,17 +368,39 @@ def job_transcode(talk_id: int, loud_key: str, final_key: str | None = None) ->
             talk = db.get(Talk, talk_id)
             if not talk:
                 raise ValueError(f"Talk {talk_id} not found")
-            job = Job(talk_id=talk_id, kind="transcode", status="running")
+            job = Job(
+                talk_id=talk_id,
+                kind="transcode",
+                status="running",
+                progress_pct=0.0,
+            )
```
ViRUS-0-0, line 376: In `app/tasks.py`, `job` is created with `progress_pct=0.0`. When media duration is unavailable or `None`, intermediate progress events are skipped, but `progress_pct` remains stuck at `0.0%` for the entire transcode run rather than staying `null`. Polling clients therefore misinterpret the job as stalled at 0.0% progress instead of recognizing that progress is unknown.

Resolution: Initialize `Job` with `progress_pct=None`. The field will only be populated with numeric percentages once `_on_progress` reports calculated values.

app/schemas.py
```diff
@@ -58,6 +58,7 @@ class JobBase(BaseModel):
     kind: str
     status: str
     log_path: str | None = None
+    progress_pct: float | None = None
```
ViRUS-0-0: In `app/schemas.py`, `JobBase.progress_pct` is typed as `float | None = None` without boundary constraints. While database logic clamps or rounds values, schema models should enforce `ge=0.0, le=100.0` using Pydantic `Field` to guarantee that API serialization and deserialization reject impossible percentage values.

 Resolution: Use `Field(default=None, ge=0.0, le=100.0)` for `progress_pct`.

tests/test_tasks.py
```diff
@@ -452,6 +453,44 @@ def fake_transcode(input_path, output_path):
...
+
+    def fake_transcode(input_path, output_path, on_progress=None):
+        assert db_ctx.open_sessions == 0, "DB session was open during transcode!"
+        if on_progress:
+            on_progress(0.25)
+            progress_history.append(jobs[100].progress_pct)
+            on_progress(0.60)
+            progress_history.append(jobs[100].progress_pct)
+            on_progress(0.95)
+            progress_history.append(jobs[100].progress_pct)
```
ViRUS-0-0: In `tests/test_tasks.py`, `test_transcode_progress_callback_updates_job_progress` accesses `jobs[100]` by hardcoded integer key. If `MockDBContext` or test setup changes its primary key sequencing, the test will raise a `KeyError`. Furthermore, the test only checks `db_ctx.open_sessions == 0` once at the start of `fake_transcode` rather than verifying that `open_sessions == 0` after each `on_progress` call, failing to rigorously assert invariant.

Resolution: Dynamically determine the active `job_id` from `jobs`, and assert `db_ctx.open_sessions == 0` after every `on_progress` call.

app/pipeline/transcode.py
```diff
@@ -227,7 +227,7 @@ def transcode(
                             0.0, current_ts_s - stream_first_pts[stream_idx]
                         )
                         pct = min(0.99, max(0.0, elapsed_s / in_duration_s))
-                        if pct - last_reported_pct >= 0.02:
+                        if pct > last_reported_pct:
```
ViRUS-0-0: In `app/pipeline/transcode.py`, changing `if pct - last_reported_pct >= 0.02:` to `if pct > last_reported_pct:` causes `on_progress` to be called on virtually every packet or frame. Because `last_reported_pct` stores the raw unrounded float `pct`, but `on_progress` receives `round(pct, 4)`, whenever `pct` advances by less than `0.0001` (typical for multi-minute or hour-long streams), `on_progress` is invoked repeatedly with the exact same rounded value. This adds unnecessary function call and clock-query overhead inside the performance-critical media decode/encode loop.

ViRUS-0-0 (changes requested):

ViRUS-0-0 (changes requested):
regenerate migration

Saksham-Sirohi (approved):
Progress starts as `null` when duration is unknown, schema bounds are enforced, callback throttling compares rounded values, and the migration is now `0005`.

PR #199, opened 2026-09-07
feat(api): extend GET /jobs/{id} with progress and timing indicators

### LINKED ISSUE #184: [Phase 5] Extend `GET /jobs/{id}` with progress and timing
### Description
Surfaces the progress data produced in Issue 7 through the public API, so polling clients see real movement instead of a static `"transcoding"` status.

### Technical design
- Extend the `Job` response Pydantic schema (`schemas.py`) to include:
  - `progress_pct` (nullable float/int)
  - `started_at`, `updated_at` (existing/already-tracked timestamps if present from Phase 4; add if missing)
  - `elapsed_time` — computed field, `now - started_at`
  - `estimated_remaining` — computed only when `progress_pct > 0`; simple linear extrapolation (`elapsed_time / progress_pct * (1 - progress_pct)`) is acceptable for v1, no need for smoothing/averaging.
- Update `GET /jobs/{id}` handler to populate these fields from the `Job` row (already being written by Issue 7).
- Update OpenAPI docstrings/examples to reflect the new fields.

### Scope
- Response schema extension.
- Handler changes to populate and return the new fields.
- Docs/OpenAPI update.

### Out of scope
- Any push-based progress mechanism (websockets, SSE) — polling only, per existing API pattern. Push notification is a Phase 7 webhook concern, and even then scoped to state changes, not granular progress.
- Progress display in any UI beyond the minimal ops view (not required by this issue).

### Outcome
Clients polling `GET /jobs/{id}` during a transcode see a numeric progress percentage and time estimates that move over successive polls, rather than a static status string.

### Acceptance Criteria
- Polling `GET /jobs/{id}` during a real transcode shows `progress_pct` changing between polls.
- `elapsed_time` is always present once a job has `started_at` set; `estimated_remaining` appears only once `progress_pct > 0` and is a sane (non-negative, non-absurd) value.
- OpenAPI schema reflects the new fields; existing clients ignoring them see no breaking change (additive only).

### Dependency
Depends on **Issue 7** (needs the underlying data being written). Independent of Issues 1–6.

### LINKED ISSUE #198: [UI] Display real-time progress percentage and timing estimates in Studio & Dashboard
### Description
Following the implementation of the Studio & Dashboard interfaces (#186) and the transcode progress tracking pipeline (#189), the UI should surface real-time job progress percentages, elapsed time, and estimated remaining time rather than only static status badges.

### Scope & Technical Design
- **Studio Sidebar (`studio.html` & `studio.js`)**:
  - In the "Recent Jobs" panel, display the progress percentage (e.g. `transcode (45%)`) and a mini progress bar for active jobs.
  - Show computed `elapsed_time` and `estimated_remaining` duration badges next to running/completed pipeline jobs.
- **Conference Dashboard (`dashboard.html` & `dashboard.js`)**:
  - When talks are in active stages (e.g., `transcoding`), update the polling logic to query job progress and display granular progress percentages in the status column.

### Dependencies
- Depends on #184 (`[Phase 5] Extend GET /jobs/{id} with progress and timing`).

### Outcome
Operators and reviewers monitoring talks in the Studio or Dashboard see real-time progress bars, percentage indicators, and accurate countdown/elapsed estimates.

SxxAq:
### Description
Extends `GET /jobs/{id}` with lifecycle timestamps (`started_at`, `updated_at`), computed elapsed execution time, and linearly extrapolated estimated remaining time for video pipeline operations. Additionally surfaces real-time progress percentages, mini progress bars, elapsed time, and countdowns in the Review Studio UI.

image

Closes #184
Closes #198

---

### Key Changes
- **Database & Migrations**:
  - Added migration `0006_add_job_timing_columns.py` adding `started_at` and `updated_at` (`DateTime(timezone=True)`, nullable) columns to table `jobs`.
  - Updated `Job` model in `app/models.py` with timestamps and helper properties for `elapsed_time` and `estimated_remaining`.
- **Schemas & API**:
  - Extended `JobBase` / `JobRead` with `started_at`, `updated_at`, `@computed_field elapsed_time`, and `@computed_field estimated_remaining` (linearly extrapolated when `progress_pct > 0`).
  - Updated `GET /jobs/{id}` endpoint docstring and OpenAPI specs.
- **Pipeline & Tasks**:
  - Updated `app/tasks.py` and `app/routes/ui.py` to record `started_at` on job creation and refresh `updated_at` on progress ticks and completion/failure.
- **Studio UI Enhancements**:
  - Updated the **Recent Jobs** panel in `studio.html` to display progress percentage badges (`45%`), animated mini progress bars, start timestamps, elapsed duration, and estimated remaining countdowns.
  - Added interactive animated pipeline progress bar in Review Actions with dynamic stage updates upon approving timestamps or previews.
- **Testing**:
  - Added test coverage in `tests/test_jobs.py` for mathematical accuracy of linear extrapolation, boundary conditions (0%, 100%, null start), and timezone handling.
  - Hardened fixture cleanup in `tests/test_ui.py`.

---

### Verification
- `uv run pytest tests/test_jobs.py tests/test_ui.py` passed 19/19 tests.
- Full test suite passed (321 passed, 1 deselected).
- `uv run ruff check .` and `uv run ruff format --check .` passed cleanly.

## Summary by Sourcery

Expose live job progress and timing information across the API and Review Studio.

New Features:
- Expose job lifecycle timestamps, elapsed execution time, and estimated remaining time through job and talk APIs.
- Display live pipeline progress, timing details, progress bars, and countdowns in the dashboard and Review Studio.
- Add an authenticated endpoint for polling recent jobs and active talk progress.

Enhancements:
- Record job start, progress, completion, and failure timing consistently across pipeline tasks.
- Expand talk responses to include associated job information and active pipeline status handling.

Tests:
- Add coverage for progress estimation, terminal-job timing, timezone and boundary behavior, job polling authorization, and UI progress rendering.

Chores:
- Add database support for job timing metadata.

## Summary by CodeRabbit

- **New Features**
  - Added live processing progress indicators with percentages, elapsed time, and estimated time remaining.
  - Added recent job history to talk details, including status, timing, and progress updates.
  - Added progress displays to the dashboard and studio views for active processing stages.
  - Expanded status indicators for additional processing states, including rejected and broken jobs.

- **Bug Fixes**
  - Improved job timing and status updates during active, completed, and failed processing.
  - Improved dashboard polling so progress remains visible with or without an API key.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -83,8 +83,6 @@ class Job(Base):
     progress_pct: Mapped[float | None] = mapped_column(Float, nullable=True)
     started_at: Mapped[datetime | None] = mapped_column(
         DateTime(timezone=True),
-        server_default=func.now(),
-        default=lambda: datetime.now(UTC),
         nullable=True,
     )
     updated_at: Mapped[datetime | None] = mapped_column(
@@ -101,8 +99,8 @@ class Job(Base):
     def elapsed_time(self) -> float | None:
         """Calculate elapsed runtime in seconds.
 
-        Uses `updated_at` as the end time for terminal statuses (`done`, `failed`)
-        when present, falling back to current UTC time for non-terminal jobs.
+        Uses `updated_at` as the end time for terminal statuses (`done`, `failed`, `broken`, `rejected`)
+        when present, returning None if `updated_at` is missing for terminal jobs to prevent time drift.
         """
         if self.started_at is None:
             return None
@@ -111,7 +109,9 @@ class Job(Base):
             if self.started_at.tzinfo is not None
             else self.started_at.replace(tzinfo=UTC)
         )
-        if self.status in ("done", "failed") and self.updated_at is not None:
+        if self.status in ("done", "failed", "broken", "rejected"):
+            if self.updated_at is None:
+                return None
             end = (
                 self.updated_at
                 if self.updated_at.tzinfo is not None
@@ -124,6 +124,8 @@ class Job(Base):
     @property
     def estimated_remaining(self) -> float | None:
         """Estimate remaining runtime in seconds based on elapsed time and progress percentage."""
+        if self.status != "running":
+            return 0.0 if self.status == "done" and self.progress_pct is not None and self.progress_pct >= 100.0 else None
         if (
             self.elapsed_time is None
             or self.progress_pct is None
--- a/app/models.py
+++ b/app/models.py
@@ -125,7 +125,13 @@ class Job(Base):
     def estimated_remaining(self) -> float | None:
         """Estimate remaining runtime in seconds based on elapsed time and progress percentage."""
         if self.status != "running":
-            return 0.0 if self.status == "done" and self.progress_pct is not None and self.progress_pct >= 100.0 else None
+            return (
+                0.0
+                if self.status == "done"
+                and self.progress_pct is not None
+                and self.progress_pct >= 100.0
+                else None
+            )
         if (
             self.elapsed_time is None
             or self.progress_pct is None
--- a/app/models.py
+++ b/app/models.py
@@ -81,8 +81,47 @@ class Job(Base):
     status: Mapped[str] = mapped_column(String(50), nullable=False)
     log_path: Mapped[str | None] = mapped_column(Text)
     progress_pct: Mapped[float | None] = mapped_column(Float, nullable=True)
+    started_at: Mapped[datetime | None] = mapped_column(
+        DateTime(timezone=True),
+        server_default=func.now(),
+        default=lambda: datetime.now(UTC),
+        nullable=True,
+    )
+    updated_at: Mapped[datetime | None] = mapped_column(
+        DateTime(timezone=True),
+        server_default=func.now(),
+        default=lambda: datetime.now(UTC),
+        onupdate=lambda: datetime.now(UTC),
+        nullable=True,
+    )
 
     talk: Mapped[Talk] = relationship(back_populates="jobs")
+
+    @property
+    def elapsed_time(self) -> float | None:
+        if self.started_at is None:
+            return None
+        now = datetime.now(UTC)
+        started = (
+            self.started_at
+            if self.started_at.tzinfo is not None
+            else self.started_at.replace(tzinfo=UTC)
+        )
+        return max(0.0, round((now - started).total_seconds(), 2))
+
+    @property
+    def estimated_remaining(self) -> float | None:
+        if (
+            self.elapsed_time is None
+            or self.progress_pct is None
+            or self.progress_pct <= 0.0
+        ):
+            return None
+        if self.progress_pct >= 100.0:
+            return 0.0
+        pct_fraction = self.progress_pct / 100.0
+        remaining = (self.elapsed_time / pct_fraction) * (1.0 - pct_fraction)
+        return max(0.0, round(remaining, 2))
 
 
 class Review(Base):
--- a/app/models.py
+++ b/app/models.py
@@ -99,18 +99,31 @@ class Job(Base):
 
     @property
     def elapsed_time(self) -> float | None:
+        """Calculate elapsed runtime in seconds.
+
+        Uses `updated_at` as the end time for terminal statuses (`done`, `failed`)
+        when present, falling back to current UTC time for non-terminal jobs.
+        """
         if self.started_at is None:
             return None
-        now = datetime.now(UTC)
         started = (
             self.started_at
             if self.started_at.tzinfo is not None
             else self.started_at.replace(tzinfo=UTC)
         )
-        return max(0.0, round((now - started).total_seconds(), 2))
+        if self.status in ("done", "failed") and self.updated_at is not None:
+            end = (
+                self.updated_at
+                if self.updated_at.tzinfo is not None
+                else self.updated_at.replace(tzinfo=UTC)
+            )
+        else:
+            end = datetime.now(UTC)
+        return max(0.0, round((end - started).total_seconds(), 2))
 
     @property
     def estimated_remaining(self) -> float | None:
+        """Estimate remaining runtime in seconds based on elapsed time and progress percentage."""
         if (
             self.elapsed_time is None
             or self.progress_pct is None
--- a/app/models.py
+++ b/app/models.py
@@ -109,7 +109,7 @@ class Job(Base):
             if self.started_at.tzinfo is not None
             else self.started_at.replace(tzinfo=UTC)
         )
-        if self.status in ("done", "failed", "broken", "rejected"):
+        if self.status in ("done", "failed", "broken", "rejected", "cancelled"):
             if self.updated_at is None:
                 return None
             end = (
--- a/app/routes/ui.py
+++ b/app/routes/ui.py
@@ -549,13 +549,13 @@ def _get_scoped_talk(talk_id: int, client: models.Client, db: Session) -> models
     return talk
 
 
-@router.get("/talks/{talk_id}/jobs", response_model=list[schemas.JobRead])
+@router.get("/talks/{talk_id}/jobs", response_model=schemas.TalkJobsResponse)
 def get_talk_jobs(
     talk_id: int,
     db: Annotated[Session, Depends(get_db)],
     client: Annotated[models.Client | None, Depends(get_optional_ui_client)] = None,
 ):
-    """Returns the recent jobs and active progress for a talk."""
+    """Returns the current talk status along with recent jobs and active progress."""
     talk = db.query(models.Talk).filter(models.Talk.id == talk_id).first()
     if not talk or (client is not None and talk.event_id not in client.event_ids):
         raise HTTPException(status_code=404, detail="Talk not found")
@@ -566,7 +566,7 @@ def get_talk_jobs(
         .limit(10)
         .all()
     )
-    return jobs
+    return {"status": talk.status, "jobs": jobs}
 
 
 @router.post("/talks/{talk_id}/status")
--- a/app/routes/ui.py
+++ b/app/routes/ui.py
@@ -253,6 +253,9 @@ def _execute_full_processing_pipeline(
             kind=kind,
             status="done",
             log_path=f"{talk_id}/logs/{kind}.log",
+            progress_pct=100.0,
+            started_at=datetime.now(UTC),
+            updated_at=datetime.now(UTC),
         )
         db.add(job)
 
@@ -293,6 +296,9 @@ def _execute_master_transcode_pipeline(
             kind=kind,
             status="done",
             log_path=f"{talk_id}/logs/{kind}.log",
+            progress_pct=100.0,
+            started_at=datetime.now(UTC),
+            updated_at=datetime.now(UTC),
         )
         db.add(job)
 
--- a/app/routes/ui.py
+++ b/app/routes/ui.py
@@ -19,7 +19,7 @@ from fastapi import (
 from fastapi.responses import FileResponse, HTMLResponse
 from fastapi.templating import Jinja2Templates
 from pydantic import BaseModel
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session, selectinload
 
 from app import models, schemas
 from app.auth import hash_api_key
@@ -189,6 +189,7 @@ def _record_stage_job(
     db.add(job)
     db.commit()
     db.refresh(job)
+    job_id = job.id
     try:
         result = func()
         job.status = "done"
@@ -197,9 +198,12 @@ def _record_stage_job(
         db.commit()
         return result
     except Exception:
-        job.status = "failed"
-        job.updated_at = datetime.now(UTC)
-        db.commit()
+        db.rollback()
+        failed_job = db.get(models.Job, job_id)
+        if failed_job:
+            failed_job.status = "failed"
+            failed_job.updated_at = datetime.now(UTC)
+            db.commit()
         raise
 
 
@@ -344,7 +348,7 @@ def dashboard(
     status_filter: str | None = None,
     q: str | None = None,
 ):
-    query = db.query(models.Talk)
+    query = db.query(models.Talk).options(selectinload(models.Talk.jobs))
     if client is not None:
         query = query.filter(models.Talk.event_id.in_(client.event_ids))
     if event_id is not None:
@@ -552,13 +556,11 @@ def _get_scoped_talk(talk_id: int, client: models.Client, db: Session) -> models
 @router.get("/talks/{talk_id}/jobs", response_model=schemas.TalkJobsResponse)
 def get_talk_jobs(
     talk_id: int,
+    client: Annotated[models.Client, Depends(get_ui_client)],
     db: Annotated[Session, Depends(get_db)],
-    client: Annotated[models.Client | None, Depends(get_optional_ui_client)] = None,
 ):
     """Returns the current talk status along with recent jobs and active progress."""
-    talk = db.query(models.Talk).filter(models.Talk.id == talk_id).first()
-    if not talk or (client is not None and talk.event_id not in client.event_ids):
-        raise HTTPException(status_code=404, detail="Talk not found")
+    talk = _get_scoped_talk(talk_id, client, db)
     jobs = (
         db.query(models.Job)
         .filter(models.Job.talk_id == talk_id)
--- a/app/routes/ui.py
+++ b/app/routes/ui.py
@@ -20,7 +20,7 @@ from fastapi.templating import Jinja2Templates
 from pydantic import BaseModel
 from sqlalchemy.orm import Session
 
-from app import models
+from app import models, schemas
 from app.auth import hash_api_key
 from app.config import PREVIEW_PRESETS
 from app.db import get_db
@@ -515,6 +515,26 @@ def _get_scoped_talk(talk_id: int, client: models.Client, db: Session) -> models
             detail="Talk not found",
         )
     return talk
+
+
+@router.get("/talks/{talk_id}/jobs", response_model=list[schemas.JobRead])
+def get_talk_jobs(
+    talk_id: int,
+    db: Annotated[Session, Depends(get_db)],
+    client: Annotated[models.Client | None, Depends(get_optional_ui_client)] = None,
+):
+    """Returns the recent jobs and active progress for a talk."""
+    talk = db.query(models.Talk).filter(models.Talk.id == talk_id).first()
+    if not talk or (client is not None and talk.event_id not in client.event_ids):
+        raise HTTPException(status_code=404, detail="Talk not found")
+    jobs = (
+        db.query(models.Job)
+        .filter(models.Job.talk_id == talk_id)
+        .order_by(models.Job.id.desc())
+        .limit(10)
+        .all()
+    )
+    return jobs
 
 
 @router.post("/talks/{talk_id}/status")
--- a/app/routes/ui.py
+++ b/app/routes/ui.py
@@ -1,8 +1,9 @@
 import json
 import tempfile
+from collections.abc import Callable
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
-from typing import Annotated
+from typing import Annotated, Any
 
 import jinja2
 from fastapi import (
@@ -165,6 +166,43 @@ def get_evaluated_milestones(status: str) -> list[dict]:
     return result
 
 
+def _record_stage_job(
+    talk_id: int,
+    kind: str,
+    db: Session,
+    func: Callable[[], Any],
+) -> Any:
+    """Execute a pipeline stage function while recording running and completed Job timestamps and status.
+
+    Creates and commits a running Job record before running `func`, and updates
+    it to done (with 100% progress) or failed upon completion or exception.
+    """
+    job = models.Job(
+        talk_id=talk_id,
+        kind=kind,
+        status="running",
+        log_path=f"{talk_id}/logs/{kind}.log",
+        progress_pct=0.0,
+        started_at=datetime.now(UTC),
+        updated_at=datetime.now(UTC),
+    )
+    db.add(job)
+    db.commit()
+    db.refresh(job)
+    try:
+        result = func()
+        job.status = "done"
+        job.progress_pct = 100.0
+        job.updated_at = datetime.now(UTC)
+        db.commit()
+        return result
+    except Exception:
+        job.status = "failed"
+        job.updated_at = datetime.now(UTC)
+        db.commit()
+        raise
+
+
 def _execute_full_processing_pipeline(
     talk: models.Talk,
     storage: StorageBackend,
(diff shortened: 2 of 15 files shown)


migrations/versions/0007_add_job_timing_columns.py
ViRUS-0-0, line 1: Regenerate migration

ViRUS-0-0 (changes requested):
Please resolve conflicts and also resolve the comments left by ai.

Moreover also regenerate the migration

app/routes/ui.py
```diff
@@ -511,6 +549,26 @@ def _get_scoped_talk(talk_id: int, client: models.Client, db: Session) -> models
...
 
+@router.get("/talks/{talk_id}/jobs", response_model=schemas.TalkJobsResponse)
+def get_talk_jobs(
+    talk_id: int,
+    db: Annotated[Session, Depends(get_db)],
+    client: Annotated[models.Client | None, Depends(get_optional_ui_client)] = None,
+):
+    """Returns the current talk status along with recent jobs and active progress."""
+    talk = db.query(models.Talk).filter(models.Talk.id == talk_id).first()
+    if not talk or (client is not None and talk.event_id not in client.event_ids):
```
ViRUS-0-0: In `app/routes/ui.py`, the endpoint `get_talk_jobs` uses `get_optional_ui_client`. The authorization check `if not talk or (client is not None and talk.event_id not in client.event_ids)` only validates event ownership if `client` is NOT None. When an unauthenticated caller makes a request without an API key, `client` evaluates to `None`, bypassing the tenant scoping check entirely and allowing arbitrary users to view talk status and job history across all events.

Resolution: Require authentication via `get_client` if this endpoint handles tenant-scoped data, or verify that unauthenticated access is strictly rejected if talk scoping applies to the environment.

app/models.py
```diff
@@ -81,9 +81,61 @@ class Job(Base):
...
+            self.elapsed_time is None
+            or self.progress_pct is None
+            or self.progress_pct <= 0.0
+        ):
+            return None
+        if self.progress_pct >= 100.0:
+            return 0.0
+        pct_fraction = self.progress_pct / 100.0
+        remaining = (self.elapsed_time / pct_fraction) * (1.0 - pct_fraction)
+        return max(0.0, round(remaining, 2))
```
ViRUS-0-0, line 145: Both `models.Job.estimated_remaining` and `schemas.JobRead.estimated_remaining` calculate remaining time solely using `progress_pct` without checking `self.status`. If a job fails at 40% progress, it reports an estimated remaining time as if it were still running. This bug was even codified into `tests/test_jobs.py` line 286, which asserts `failed_job.estimated_remaining == 225.0`. An aborted or failed job must never report a remaining countdown.

Resolution: Check `self.status` in `estimated_remaining` across both `models.py` and `schemas.py`. Return `None` if the job is not in an active running state (`self.status != 'running'`), or `0.0` if `self.status == 'done'`.

app/schemas.py
```diff
@@ -82,6 +91,51 @@ class JobRead(JobBase):
...
+            else self.started_at.replace(tzinfo=UTC)
+        )
+        if self.status in ("done", "failed") and self.updated_at is not None:
+            end = (
+                self.updated_at
+                if self.updated_at.tzinfo is not None
+                else self.updated_at.replace(tzinfo=UTC)
+            )
+        else:
+            end = datetime.now(UTC)
```
ViRUS-0-0: In `models.Job.elapsed_time` and `schemas.JobRead.elapsed_time`, the code checks `if self.status in ("done", "failed") and self.updated_at is not None:`. If a terminal job has `updated_at == None` (which is true for all legacy jobs prior to migration 0007 or incomplete executions), execution falls into the `else:` branch, setting `end = datetime.now(UTC)`. Consequently, a completed or failed historical job's elapsed time continuously increases with current real time.

 Resolution: If the job is in a terminal status (`done`, `failed`, `rejected`, `broken`) but lacks an `updated_at` timestamp, return `None` because the true execution runtime cannot be determined.

app/models.py
```diff
@@ -81,9 +81,61 @@ class Job(Base):
     status: Mapped[str] = mapped_column(String(50), nullable=False)
     log_path: Mapped[str | None] = mapped_column(Text)
     progress_pct: Mapped[float | None] = mapped_column(Float, nullable=True)
+    started_at: Mapped[datetime | None] = mapped_column(
+        DateTime(timezone=True),
+        server_default=func.now(),
+        default=lambda: datetime.now(UTC),
+        nullable=True,
+    )
```
ViRUS-0-0, line 87: In `app/models.py`, `Job.started_at` is defined with `server_default=func.now()` and `default=lambda: datetime.now(UTC)`. When a job is enqueued (e.g. `Job(status='queued')`), SQLAlchemy sets `started_at` to the insertion timestamp. Because `started_at` is present, `elapsed_time` computes queue latency as execution duration, violating Issue #184 acceptance criteria stating that `started_at` is only set once execution begins.

 Resolution: Remove `server_default` and `default` from `started_at` in `models.Job`. It should default to `None` and only be set when the worker or pipeline actively transitions the job to `running`.

app/tasks.py
```diff
@@ -59,6 +60,7 @@ def _handle_failure(talk_id: int, job_id: int | None, exc: Exception, storage) -
                 return
             job.status = "failed"
             job.log_path = log_key
+            job.updated_at = datetime.now(UTC)
```
ViRUS-0-0, line 63: In `app/tasks.py`, the only updates to timing columns are in `_handle_failure` and `_on_progress`. When a background worker takes a queued job and starts transcoding, it never populates `started_at`. Furthermore, when a job finishes successfully (`status = 'done'`), `updated_at` is not set to `datetime.now(UTC)`. This leaves `updated_at` at the timestamp of the last progress event, ignoring post-processing runtime.

Resolution: Ensure `app/tasks.py` sets `job.started_at = datetime.now(UTC)` and `job.updated_at = datetime.now(UTC)` when execution begins, and updates `job.updated_at = datetime.now(UTC)` when setting `job.status = 'done'` upon completion.

app/routes/ui.py
```diff
@@ -165,6 +166,43 @@ def get_evaluated_milestones(status: str) -> list[dict]:
...
+        job.status = "done"
+        job.progress_pct = 100.0
+        job.updated_at = datetime.now(UTC)
+        db.commit()
+        return result
+    except Exception:
+        job.status = "failed"
+        job.updated_at = datetime.now(UTC)
+        db.commit()
+        raise
```
ViRUS-0-0, line 207: In `app/routes/ui.py:198`, `_record_stage_job` catches exceptions from `func()` and attempts `job.status = 'failed'; job.updated_at = ...; db.commit()`. If `func()` fails due to an internal database or SQLAlchemy error, the transaction session is invalid. Attempting `db.commit()` without calling `db.rollback()` first raises `sqlalchemy.exc.PendingRollbackError`, masking the underlying exception and failing to persist the failed job record.

Resolution: Call `db.rollback()` before attempting to update the job status to 'failed', and re-query or merge the job instance before committing.

app/ui/static/js/dashboard.js
```diff
@@ -26,22 +32,92 @@ function getActiveTalkIds() {
...
+      renderActiveJobCell(cell, talkStatus || activeJob.kind, pct, remainingStr);
+    } else if (talkStatus) {
+      const newBadge = STATUS_BADGE_MAP[talkStatus] ?? STATUS_BADGE_MAP.waiting_for_files;
+      cell.innerHTML = newBadge;
+    }
+
+    if (row && talkStatus) row.dataset.status = talkStatus;
+    if (talkStatus === 'done' || talkStatus === 'failed') {
       setTimeout(() => location.reload(), 1500);
     }
```
ViRUS-0-0: In `app/ui/static/js/dashboard.js`, when any polled talk reaches status `'done'` or `'failed'`, line 119 triggers `setTimeout(() => location.reload(), 1500)`. This destroys active user interactions, such as search input filters, scroll position, and talk selections on the dashboard.

Resolution: Update the table cell badge and row dataset in place rather than forcing a full page reload.

app/ui/static/js/dashboard.js
```diff
@@ -26,22 +32,92 @@ function getActiveTalkIds() {
...
+      jobs = data.jobs || [];
+    } else {
+      const r = await fetch(`/studio/talks/${talkId}/jobs`);
+      if (!r.ok) return;
+      const data = await r.json();
+      talkStatus = data.status || talkStatus;
+      jobs = data.jobs || (Array.isArray(data) ? data : []);
+    }
+
+    const activeJob = jobs.find(j => j.status === 'running');
```
ViRUS-0-0: In `app/ui/static/js/dashboard.js`, `jobs.find(j => j.status === 'running')` is used to find the active job. When jobs are returned in default primary key ascending order (e.g. from `TalkWithJobsRead`), `find` returns the oldest job. If a prior job stalled or terminated uncleanly in 'running' state, the dashboard will continuously report the dead job instead of the latest active stage.

Resolution: Sort jobs descending by `id` or reverse before finding the running job so the most recent active stage is prioritized.

app/ui/templates/dashboard.html
```diff
@@ -167,7 +167,25 @@
             {% else %}—{% endif %}
           </td>
           <td class="status-cell" data-talk-id="{{ talk.id }}">
-            {{ status_badge(talk.status) }}
+            {% set active_job = (talk.jobs | selectattr('status', 'equalto', 'running') | list | first) if talk.jobs else none %}
```
ViRUS-0-0, line 170: In `app/ui/templates/dashboard.html:170`, the template checks `{% set active_job = (talk.jobs | selectattr('status', 'equalto', 'running') | list | first) if talk.jobs else none %}`. Because `talks` in `app/routes/ui.py` is queried without `selectinload(models.Talk.jobs)`, accessing `talk.jobs` inside the loop causes a separate SQL query for every talk displayed in the dashboard table.

 Resolution: Update the dashboard database query in `app/routes/ui.py` to eagerly load `jobs` using `options(selectinload(models.Talk.jobs))`.

app/ui/static/js/studio.js
```diff
@@ -446,10 +482,131 @@ window.handleVideoFileUpload = async function(e, talkId) {
...
+    const data = await res.json();
+    const jobs = Array.isArray(data) ? data : (data.jobs || []);
+    renderRecentJobs(jobs);
+  } catch { /* skip */ }
+}
+
+function startStudioPolling() {
+  if (studioPollInterval) clearInterval(studioPollInterval);
+  studioPollInterval = setInterval(pollStudioJobs, 2500);
+}
```
ViRUS-0-0, line 595: In `app/ui/static/js/studio.js:597`, `setInterval(pollStudioJobs, 2500)` starts on page load and never ceases. Even if a talk has no running jobs or is in a terminal state (`done`/`failed`), polling requests to `/studio/talks/{talkId}/jobs` continue every 2.5 seconds indefinitely.

Resolution: Check if any returned job is in `status === 'running'`. If no jobs are running and the talk is in a terminal status, stop or back off the polling interval.

app/ui/static/css/app.css
```diff
@@ -544,6 +544,55 @@ td {
...
+  background: var(--v-primary);
+  border-radius: 2px;
+  transition: width 0.4s cubic-bezier(0.4, 0, 0.2, 1);
+}
+
+.job-progress-fill.animated {
+  background: linear-gradient(90deg, var(--v-primary) 0%, #38bdf8 50%, var(--v-primary) 100%);
+  background-size: 200% 100%;
+  animation: progress-shimmer 1.5s infinite linear;
+}
```
ViRUS-0-0, line 580: In `app/ui/static/css/app.css:575`, `.job-progress-fill.animated` runs `animation: progress-shimmer 1.5s infinite linear;`. Continuous animations without a reduced-motion fallback violate accessibility standards (WCAG 2.2.2 and 2.3.3) for users with vestibular disorders.

 Resolution: Add a `@media (prefers-reduced-motion: reduce)` media query disabling the animation.

app/ui/templates/studio.html
```diff
@@ -348,18 +363,40 @@
...
+            </div>
+          </div>
+          {% if job.status == 'running' and job.progress_pct is not none %}
+          <div class="job-progress-track">
+            <div class="job-progress-fill" style="width: {{ job.progress_pct }}%;"></div>
+          </div>
+          {% endif %}
+          {% if job.started_at %}
+          <div class="job-timing-meta">
+            <span>Started {{ job.started_at.strftime('%H:%M:%S') }}</span>
```
ViRUS-0-0: In `app/ui/templates/studio.html:387`, `job.started_at.strftime('%H:%M:%S')` formats the raw UTC timestamp during initial server render. When client-side polling in `app/ui/static/js/studio.js:560` fires, `new Date(job.started_at).toTimeString()` formats the date in the user's local timezone. This creates an unexpected visual time jump after the first polling cycle.

Resolution: Append UTC to the server rendered timestamp (e.g. `strftime('%H:%M:%S UTC')`) or standardize client-side formatting.

ViRUS-0-0 (changes requested):
Please take a look at comments.
I have provided a checklist please see to it

- [ ] Fix authorization bypass in `app/routes/ui.py:559` by enforcing authentication or scoping checks when `client is None`
- [ ] Restrict `estimated_remaining` to active non-terminal jobs (`status == 'running'`) in `app/models.py:124` and `app/schemas.py:119`
- [ ] Correct test assertion in `tests/test_jobs.py:286` to expect `estimated_remaining is None` for failed jobs
- [ ] Prevent elapsed time drift on terminal jobs with missing `updated_at` by returning `None` instead of `datetime.now(UTC)` in `app/models.py:113` and `app/schemas.py:108`
- [ ] Remove `server_default` and `default=lambda: datetime.now(UTC)` from `Job.started_at` in `app/models.py:84` so queued jobs do not record start times prematurely
- [ ] Ensure `app/tasks.py` stamps `job.started_at` on task start and updates `job.updated_at` upon successful completion (`status == 'done'`)
- [ ] Add `db.rollback()` before updating failed status in `_record_stage_job` in `app/routes/ui.py:198` to prevent `PendingRollbackError`
- [ ] Avoid disruptive full page reload (`location.reload()`) on talk completion in `app/ui/static/js/dashboard.js:119`
- [ ] Select latest running job instead of oldest running job (`jobs.find`) in `app/ui/static/js/dashboard.js:104`
- [ ] Add termination condition or polling backoff in `app/ui/static/js/studio.js:597` when all jobs are done
- [ ] Eagerly load `talk.jobs` for dashboard view in `app/routes/ui.py` to eliminate N+1 queries in `app/ui/templates/dashboard.html:170`
- [ ] Add `@media (prefers-reduced-motion: reduce)` rule for `.job-progress-fill.animated` in `app/ui/static/css/app.css:575`

app/routes/ui.py
Saksham-Sirohi: **Security:** when `client is None`, this only checks talk existence, so any anonymous caller can poll job status/log paths/timing for arbitrary `talk_id`. Require the same auth/scoping as other studio mutations (reject unauthenticated, or always scope via session/API key). The new unauthenticated test also locks in the hole.

app/models.py
```diff
@@ -81,9 +81,61 @@ class Job(Base):
     status: Mapped[str] = mapped_column(String(50), nullable=False)
     log_path: Mapped[str | None] = mapped_column(Text)
     progress_pct: Mapped[float | None] = mapped_column(Float, nullable=True)
+    started_at: Mapped[datetime | None] = mapped_column(
+        DateTime(timezone=True),
+        server_default=func.now(),
+        default=lambda: datetime.now(UTC),
```
Saksham-Sirohi: Drop `server_default=func.now()` and `default=lambda: datetime.now(UTC)` from `started_at`. Queued/created rows must not look “started”. Stamp `started_at` explicitly when a job enters `running` (UI `_record_stage_job` already does; `app/tasks.py` must too). Migration 0007 correctly adds nullable columns without a DB default—keep model and migration aligned.

app/models.py
```diff
@@ -81,9 +81,61 @@ class Job(Base):
...
+                self.updated_at
+                if self.updated_at.tzinfo is not None
+                else self.updated_at.replace(tzinfo=UTC)
+            )
+        else:
+            end = datetime.now(UTC)
+        return max(0.0, round((end - started).total_seconds(), 2))
+
+    @property
+    def estimated_remaining(self) -> float | None:
```
Saksham-Sirohi, line 125: `estimated_remaining` must return `None` unless `status == "running"` (and progress is usable). Failed/done jobs should never advertise remaining time. Mirror the same guard in `schemas.JobRead.estimated_remaining`.

app/schemas.py
Saksham-Sirohi, line 118: For terminal statuses (`done`/`failed`), if `updated_at` is missing, return `None` for `elapsed_time` instead of `datetime.now(UTC)`. Using “now” makes completed durations drift on every poll. Same fix in `models.Job.elapsed_time` (the `else: end = datetime.now(UTC)` branch).

app/tasks.py
Saksham-Sirohi, line 63: Good to set `updated_at` on failure—but worker paths still create `Job(status="running")` without stamping `started_at`, and success paths that set `status="done"` often omit `updated_at`. Stamp both on start; bump `updated_at` on progress (line ~560) and on successful completion for every stage job.

app/routes/ui.py
```diff
@@ -165,6 +166,43 @@ def get_evaluated_milestones(status: str) -> list[dict]:
...
+    db.commit()
+    db.refresh(job)
+    try:
+        result = func()
+        job.status = "done"
+        job.progress_pct = 100.0
+        job.updated_at = datetime.now(UTC)
+        db.commit()
+        return result
+    except Exception:
```
Saksham-Sirohi, line 200: If `func()` fails after putting the session in a failed transaction state, `job.status = "failed"; db.commit()` can raise `PendingRollbackError`. `db.rollback()` first, then re-load/update the job (or use a fresh transaction) before re-raising.

app/ui/static/js/dashboard.js
```diff
@@ -26,22 +32,92 @@ function getActiveTalkIds() {
...
+
+      renderActiveJobCell(cell, talkStatus || activeJob.kind, pct, remainingStr);
+    } else if (talkStatus) {
+      const newBadge = STATUS_BADGE_MAP[talkStatus] ?? STATUS_BADGE_MAP.waiting_for_files;
+      cell.innerHTML = newBadge;
+    }
+
+    if (row && talkStatus) row.dataset.status = talkStatus;
+    if (talkStatus === 'done' || talkStatus === 'failed') {
       setTimeout(() => location.reload(), 1500);
```
Saksham-Sirohi: Avoid `location.reload()` here—it wipes filters/scroll and fights concurrent rows. Swap the badge via `STATUS_BADGE_MAP`, clear active progress UI, and stop polling that talk. Also treat `broken`/`rejected` as terminal alongside `done`/`failed`.

app/ui/static/js/dashboard.js
Saksham-Sirohi: `jobs.find(...)` returns the oldest running job when the authenticated `/talks/{id}` payload is ascending. Prefer the latest: `jobs.findLast?.(j => j.status === 'running') ?? [...jobs].reverse().find(...)` (studio jobs endpoint already returns `id.desc()`).

app/ui/static/js/studio.js
```diff
@@ -446,10 +482,131 @@ window.handleVideoFileUpload = async function(e, talkId) {
...
+    if (!res.ok) return;
+    const data = await res.json();
+    const jobs = Array.isArray(data) ? data : (data.jobs || []);
+    renderRecentJobs(jobs);
+  } catch { /* skip */ }
+}
+
+function startStudioPolling() {
+  if (studioPollInterval) clearInterval(studioPollInterval);
+  studioPollInterval = setInterval(pollStudioJobs, 2500);
```
Saksham-Sirohi, line 594: `setInterval(pollStudioJobs, 2500)` never stops. When there is no `running` job (or talk status is terminal), clear the interval (optionally with backoff while idle). Endless 2.5s polling is noisy for a core studio page.

app/ui/static/js/studio.js
```diff
@@ -364,16 +364,52 @@ window.approveTalk = async function(id) {
...
+
+  const stages = [
+    { pct: 25, desc: 'Rendering title intro and outro slates...' },
+    { pct: 55, desc: 'Cutting media clip to specified In/Out points...' },
+    { pct: 80, desc: 'Generating video preview & normalizing audio...' },
+    { pct: 95, desc: 'Finalizing pipeline and saving assets...' }
+  ];
+
+  let stageIndex = 0;
+  const timer = setInterval(() => {
```
Saksham-Sirohi: Approve UI advances a hardcoded stage timer unrelated to server `progress_pct`. That misrepresents the new timing API. Drive this bar from `/studio/talks/{id}/jobs` (or drop the fake stages and show indeterminate + live job cards only).

app/ui/templates/dashboard.html
Saksham-Sirohi, line 170: This reads `talk.jobs` per row. Ensure the dashboard query eager-loads jobs (`selectinload(models.Talk.jobs)`), otherwise this is an N+1 on every dashboard render.

app/ui/static/css/app.css
```diff
@@ -544,6 +544,55 @@ td {
...
+  height: 100%;
+  background: var(--v-primary);
+  border-radius: 2px;
+  transition: width 0.4s cubic-bezier(0.4, 0, 0.2, 1);
+}
+
+.job-progress-fill.animated {
+  background: linear-gradient(90deg, var(--v-primary) 0%, #38bdf8 50%, var(--v-primary) 100%);
+  background-size: 200% 100%;
+  animation: progress-shimmer 1.5s infinite linear;
```
Saksham-Sirohi, line 579: Add `@media (prefers-reduced-motion: reduce)` to disable `progress-shimmer` (static fill is fine). Continuous animation without a reduced-motion escape fails basic a11y for a primary status indicator.

tests/test_jobs.py
```diff
@@ -120,5 +120,191 @@ def test_get_job_success_null_log_path():
...
+        talk_id=1,
+        kind="transcode",
+        status="failed",
+        started_at=started,
+        updated_at=updated,
+        progress_pct=40.0,
+    )
+    assert failed_job.elapsed_time == 150.0
+    # Linear remaining for failed job at 40%: (150 / 0.4) * 0.6 = 225.0
+    assert failed_job.estimated_remaining == 225.0
```
Saksham-Sirohi: This assertion encodes the bug: failed jobs must not return an ETA. After gating `estimated_remaining` to `running`, expect `None` here (keep the elapsed=`150.0` check).

app/ui/templates/studio.html
Saksham-Sirohi: Server render uses UTC `strftime('%H:%M:%S')` while `studio.js` uses `Date#toTimeString()` (local). Pick one timezone policy (prefer local, or label UTC) so the first paint and polled cards don’t disagree.

Saksham-Sirohi (changes requested):
Thanks for the progress/timing surface—direction is right for a core studio path, but the prior review checklist is still open on this head (`b740372`), and a few extra issues need fixing before merge.

**Still blocking:** auth on `/studio/talks/{id}/jobs`, `started_at` defaults, ETA only for running jobs, terminal elapsed without `updated_at`, worker timing stamps, `_record_stage_job` rollback, dashboard reload/`find`, studio poll stop, dashboard N+1, reduced-motion.

**Also:** fake approve-progress UI, and the test that asserts ETA for failed jobs.

ViRUS-0-0 (approved):

Saksham-Sirohi (approved):
Re-reviewed on `7d525cf`. Prior REQUEST_CHANGES items look addressed:

- `/studio/talks/{id}/jobs` now requires `get_ui_client` + `_get_scoped_talk` (401 covered)
- `started_at` is nullable with no create-time default; workers/`_record_stage_job` stamp it on start
- terminal `elapsed_time` uses `updated_at` (or `None` if missing) — no poll drift
- `estimated_remaining` is `None` for failed/non-running (with `0.0` only for done@100%, which matches the UI `isRunning` gate)
- `_record_stage_job` rolls back before marking failed; progress path bumps `updated_at`

Solid API + Studio/dashboard polling coverage. Thanks for the follow-ups.

**Merge note:** this still mints Alembic `0007` (same as open #204 / #217). Whichever of these lands first is fine; the others need a revision bump on rebase.

PR #226, opened 2026-09-12
feat(api): add outbound webhook notifications on talk approval

### LINKED ISSUE #224: [LOGIN] Outbound Approval Webhook
**Description**

Once a talk's recording is approved inside VEditor, the client that originally registered that talk needs to be told, so it can take its own next steps — for the eventyay-veditor plugin specifically, that means notifying the talk's speakers. VEditor has no reason to know anything about who those speakers are or how to reach them; it only needs to reliably signal that approval happened, to whichever client is responsible for that event.

**Technical Design**

*Webhook registration*

- A machine client can register a single outbound URL where it wants to be notified.
- A shared secret is established at the same time as registration — used to sign every outbound payload so the receiver can verify it genuinely came from VEditor and wasn't forged or replayed.
- Registration is per-`Client`, not per-event — this matches the existing model where one instance-wide API key covers all of that client's events; there's no need for finer granularity yet.

*Trigger point*

- The notification fires at the same state transition other parts of the system already treat as "approval happened" — specifically, when a talk moves from awaiting approval into active cutting.
- No new state is introduced; this hooks into the existing transition logic rather than adding a parallel signal.

*Delivery*

- A notification job is queued (not sent inline on the request that caused the transition) so approval itself is never blocked by a slow or failing webhook receiver.
- Payload is deliberately minimal: talk ID, event ID, and a timestamp — enough for the receiving client to look up whatever it needs (e.g. speaker emails) on its own side. VEditor doesn't need to know or store anything about the receiver's data model.
- The payload is signed using the shared secret (HMAC or equivalent) and the signature is sent alongside it so the receiver can verify authenticity before trusting the contents.
- If the client hasn't registered a URL, nothing is sent — silently, no error logged as a failure.
- If delivery fails on the first attempt, it's retried exactly once. Beyond that, the system gives up quietly. A missed notification doesn't roll back or block the talk's own approval state — the two are decoupled by design.

**Scope**

- Registration mechanism for a client's notification URL and signing secret.
- Trigger hook on the approval transition.
- Signed delivery with single-retry-on-failure behavior.

**Out of Scope**

- Any notification event besides approval — no other transition (e.g. rejection, completion) triggers this mechanism.
- Any delivery guarantee beyond the single retry — no dead-letter queue, no manual replay tooling.
- Per-event notification URLs — one URL is tied to the client as a whole, not scoped finer.

**Outcome**

VEditor can reliably and verifiably tell any client that has registered interest the moment one of its talks is approved — closing the loop the eventyay-veditor plugin's speaker-notification step depends on, without VEditor taking on any responsibility for what happens after the notification is sent.

**Acceptance Criteria**

- A talk transitioning into approval triggers exactly one outbound notification attempt.
- A client with no registered URL receives nothing, and this is not treated as an error anywhere in the system.
- The receiving client can verify, using the shared secret from registration, that a notification genuinely came from VEditor.
- A failed delivery is retried exactly once, then abandoned without affecting the talk's own approval state.
- The payload contains only talk ID, event ID, and timestamp — no speaker, email, or other receiver-specific data.

ViRUS-0-0:
## Closes #224 

This pull request implements outbound webhook notifications when a conference talk recording is approved in VEditor.

### In Plain Terms
When event organizers finish reviewing a talk's video recording in VEditor and approve it, external platforms (such as the `eventyay-veditor` plugin) need to be notified immediately so they can update their systems and notify speakers. VEditor only needs to send a short, secure message: *"Talk #X in Event #Y was approved at this time"*.

### Key Capabilities
1. **Webhook Registration**:
   - Machine clients can register, inspect, and delete their outbound notification URL and shared secret via `POST /client/webhook`, `GET /client/webhook`, and `DELETE /client/webhook`.
   - If a secret is not provided during registration, a secure 32-character secret is generated automatically.
   - The CLI command `veditor admin create-client` now supports optional `--webhook-url` and `--webhook-secret` flags.
2. **Cryptographic Signing (HMAC-SHA256)**:
   - Outbound HTTP POST requests include an `X-VEditor-Signature: sha256=<hex>` header computed using the client's shared secret over canonical JSON payload bytes.
   - Disallows HTTP redirects (`3xx`) to avoid security issues (SSRF) or dropping payloads.
3. **Decoupled Asynchronous Delivery & Fault Tolerance**:
   - Notifications are dispatched through the background job queue (`light_queue`).
   - Human reviewers approving talks never experience network latency or delays waiting for external servers.
   - If delivery fails on the first attempt, it pauses briefly and retries once (2 attempts total).
   - If delivery fails completely or Redis is unreachable, errors are logged quietly without rolling back or blocking the talk approval transaction.
4. **Minimal Payload**:
   - Outbound payload contains strictly essential IDs and a timestamp:
     ```json
     {
       "talk_id": 1,
       "event_id": 10,
       "timestamp": "2026-09-13T03:00:00.000000Z"
     }
     ```
   - Rejection decisions (`decision="reject"`) never trigger webhooks.

---

### Prerequisites
Run database migrations:
   ```bash
   uv run alembic upgrade head
   ```

---

### Step 1: Start a Local Webhook Receiver

In a separate terminal, run a minimal Python server that prints incoming requests and headers:

```bash
python3 -c '
import http.server, hmac, hashlib

class Handler(http.server.BaseHTTPRequestHandler):
    def do_POST(self):
        length = int(self.headers.get("Content-Length", 0))
        body = self.rfile.read(length)
        sig = self.headers.get("X-VEditor-Signature", "")
        print("\n--- Incoming Webhook ---")
        print("Signature Header:", sig)
        print("Payload:", body.decode("utf-8"))
        self.send_response(200)
        self.end_headers()
        self.wfile.write(b"OK")

http.server.HTTPServer(("127.0.0.1", 9999), Handler).serve_forever()
'
```

---

### Step 2: Bootstrap a Client with an API Key

Create a test client associated with an event:

```bash
uv run veditor admin create-client --event-name "Webhook Demo"
```
*Note the returned API key and Client ID from the terminal output.*

---

### Step 3: Register the Webhook

Register the local receiver URL using the API key:

```bash
curl -X POST http://127.0.0.1:8000/client/webhook \
  -H "Authorization: Bearer <YOUR_API_KEY>" \
  -H "Content-Type: application/json" \
  -d '{
    "webhook_url": "http://127.0.0.1:9999/",
    "webhook_secret": "my-shared-secret-key-12345"
  }'
```
Expected response:
```json
{
  "webhook_url": "http://127.0.0.1:9999/",
  "webhook_secret": "my-shared-secret-key-12345"
}
```

Verify the webhook registration details:
```bash
curl -X GET http://127.0.0.1:8000/client/webhook \
  -H "Authorization: Bearer <YOUR_API_KEY>"
```
Expected response:
```json
{
  "webhook_url": "http://127.0.0.1:9999/",
  "has_secret": true
}
```

---

### Step 4: Trigger Talk Approval

1. Create and ingest a talk clip or advance a talk into `pending_approval` state.
2. Approve the talk:
   ```bash
   curl -X POST http://127.0.0.1:8000/talks/<TALK_ID>/approve \
     -H "Authorization: Bearer <YOUR_API_KEY>" \
     -H "Content-Type: application/json" \
     -d '{"decision": "approve"}'
   ```

**Verify the receiver output:**
Check the receiver terminal started in Step 1. It should print:
```text
--- Incoming Webhook ---
Signature Header: sha256=<computed_hex_digest>
Payload: {"event_id":<EVENT_ID>,"talk_id":<TALK_ID>,"timestamp":"..."}
```

---

### Step 5: Test Rejection Decision

Approve another talk with `decision="reject"`:
```bash
curl -X POST http://127.0.0.1:8000/talks/<ANOTHER_TALK_ID>/approve \
  -H "Authorization: Bearer <YOUR_API_KEY>" \
  -H "Content-Type: application/json" \
  -d '{"decision": "reject"}'
```
**Verify:** The webhook receiver receives no incoming requests.

---

### Step 6: Test Deletion of Webhook

Delete the webhook configuration:
```bash
curl -X DELETE http://127.0.0.1:8000/client/webhook \
  -H "Authorization: Bearer <YOUR_API_KEY>"
```
Expected response: `204 No Content`.

Check GET endpoint:
```bash
curl -X GET http://127.0.0.1:8000/client/webhook \
  -H "Authorization: Bearer <YOUR_API_KEY>"
```
Expected response:
```json
{
  "webhook_url": null,
  "has_secret": false
}
```

## Summary by Sourcery

Add resilient, signed outbound webhook notifications for configured clients when talk recordings advance to cutting.

New Features:
- Add authenticated APIs and CLI options for registering, inspecting, and deleting client outbound webhooks with automatically generated secrets.
- Send signed talk-processing notifications through the background queue with retry handling and redirect protection.

Bug Fixes:
- Keep talk processing successful when webhook discovery or queue delivery fails, and avoid notifications for rejected talks.

Enhancements:
- Persist webhook configuration on clients and protect talk state transitions with row-level locking.

Tests:
- Add coverage for webhook registration, lifecycle management, CLI validation, signing, retries, failure isolation, and notification behavior.

## Summary by CodeRabbit

* **New Features**
  * Clients can register, view, update, and remove webhook settings through the API.
  * Webhook secrets can be provided or securely generated automatically.
  * Added CLI options for configuring client webhooks.
  * Added CLI commands to create administrators, promote users, and list users.
  * Talk cut-bound submissions can send signed webhook notifications to configured clients.
  * Webhook delivery retries once after temporary failures.
* **Bug Fixes**
  * Webhook delivery failures no longer interrupt talk processing.
  * Invalid URLs, missing webhook details, and overly long secrets are rejected with clear validation.

### DIFF
--- a/app/routes/talks.py
+++ b/app/routes/talks.py
@@ -40,6 +40,7 @@ from app.tasks import (
     STAGE_CONFIG,
     dispatch_assembly,
     job_cut,
+    job_deliver_webhook,
     job_detect,
     job_ingest,
 )
@@ -284,6 +285,50 @@ def approve_talk(
 
     if decision == "reject":
         cleanup_intermediates(storage, talk_id)
+    else:
+        try:
+            candidate_clients: list[models.Client] = []
+            try:
+                candidate_clients = (
+                    db.query(models.Client)
+                    .filter(
+                        models.Client.event_ids.any(talk.event_id),
+                        models.Client.webhook_url.is_not(None),
+                    )
+                    .all()
+                )
+            except Exception:  # noqa: BLE001
+                candidate_clients = []
+
+            if (
+                not candidate_clients
+                and client
+                and getattr(client, "webhook_url", None)
+            ):
+                candidate_clients = [client]
+
+            for c in candidate_clients:
+                webhook_url = getattr(c, "webhook_url", None)
+                webhook_secret = getattr(c, "webhook_secret", None)
+                if webhook_url and webhook_secret:
+                    payload_data = {
+                        "talk_id": talk.id,
+                        "event_id": talk.event_id,
+                        "timestamp": datetime.now(UTC).isoformat(),
+                    }
+                    light_queue.enqueue(
+                        job_deliver_webhook,
+                        webhook_url,
+                        webhook_secret,
+                        payload_data,
+                        job_timeout=30,
+                    )
+        except Exception as exc:  # noqa: BLE001
+            logger.warning(
+                "Failed to dispatch webhook notification for talk %d: %s",
+                talk.id,
+                exc,
+            )
 
     return schemas.TalkRead.model_validate(talk)
 
--- a/app/cli.py
+++ b/app/cli.py
@@ -9,7 +9,13 @@ from app.auth import hash_api_key
 from app.db import SessionLocal
 
 
-def create_client(session: Session, event_name: str | None, event_id: int | None):
+def create_client(
+    session: Session,
+    event_name: str | None,
+    event_id: int | None,
+    webhook_url: str | None = None,
+    webhook_secret: str | None = None,
+):
     if not event_name and not event_id:
         print("Error: Must provide either --event-name or --event-id.")
         sys.exit(1)
@@ -37,9 +43,19 @@ def create_client(session: Session, event_name: str | None, event_id: int | None
     raw_api_key = secrets.token_urlsafe(32)
     hashed_key = hash_api_key(raw_api_key)
 
+    if webhook_secret and not webhook_url:
+        print("Error: Cannot provide --webhook-secret without --webhook-url.")
+        sys.exit(1)
+
+    secret = None
+    if webhook_url:
+        secret = webhook_secret or secrets.token_urlsafe(32)
+
     client = models.Client(
         hashed_key=hashed_key,
         event_ids=[selected_event_id],
+        webhook_url=webhook_url,
+        webhook_secret=secret,
     )
     session.add(client)
     session.commit()
@@ -47,6 +63,9 @@ def create_client(session: Session, event_name: str | None, event_id: int | None
 
     print(f"Created Client with ID {client.id}")
     print(f"API Key: {raw_api_key}")
+    if client.webhook_url:
+        print(f"Webhook URL: {client.webhook_url}")
+        print(f"Webhook Secret: {client.webhook_secret}")
     print("Store this key safely! It will not be shown again.")
 
 
@@ -70,13 +89,34 @@ def main():
     create_client_parser.add_argument(
         "--event-id", type=int, help="ID of an existing event to scope the client to"
     )
+    create_client_parser.add_argument(
+        "--webhook-url",
+        type=str,
+        default=None,
+        help="Outbound webhook notification URL",
+    )
+    create_client_parser.add_argument(
+        "--webhook-secret",
+        type=str,
+        default=None,
+        help="Shared secret for signing webhook notifications (auto-generated if omitted)",
+    )
 
     args = parser.parse_args()
 
     if args.command == "admin" and args.subcommand == "create-client":
         db = SessionLocal()
         try:
-            create_client(db, args.event_name, args.event_id)
+            if args.webhook_url is not None or args.webhook_secret is not None:
+                create_client(
+                    db,
+                    args.event_name,
+                    args.event_id,
+                    args.webhook_url,
+                    args.webhook_secret,
+                )
+            else:
+                create_client(db, args.event_name, args.event_id)
         finally:
             db.close()
 
--- a/app/main.py
+++ b/app/main.py
@@ -5,7 +5,7 @@ from fastapi.responses import RedirectResponse
 from fastapi.staticfiles import StaticFiles
 
 from app.queue import redis_conn
-from app.routes import auth, jobs, ops, reviews, studio, talks
+from app.routes import auth, client, jobs, ops, reviews, studio, talks
 
 app = FastAPI(title="VEditor API")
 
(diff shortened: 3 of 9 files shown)


app/routes/talks.py
```diff
@@ -291,6 +297,65 @@ def approve_talk(
...
+            for c in candidate_clients:
+                webhook_url = getattr(c, "webhook_url", None)
+                webhook_secret = getattr(c, "webhook_secret", None)
+                if webhook_url and webhook_secret:
+                    payload_data = {
+                        "talk_id": talk.id,
+                        "event_id": talk.event_id,
+                        "timestamp": datetime.now(UTC).isoformat(),
+                    }
+                    light_queue.enqueue(
```
Saksham-Sirohi, line 368: This fires on `pending_approval` → `pending_bounds`. #224 asks for the transition into active cutting, which is `submit_cut_bounds` (`advance(talk, "cutting")`). A receiver that notifies speakers will do that before anyone has submitted bounds, and the actual start of cutting sends nothing.

Saksham-Sirohi (changes requested):
Registration, HMAC, no-follow redirects, and the single retry are fine, and reject does not notify. The hook is on the wrong transition: `/approve` moves the talk to `pending_bounds`, not into cutting.

Saksham-Sirohi (approved):
The webhook now fires from `submit_cut_bounds` after the talk is committed as `cutting`. `/approve` no longer notifies, including reject. HMAC, single retry, and failure isolation are unchanged.

SxxAq (approved):
LGTM

Saksham-Sirohi (approved):
A failed enqueue for one webhook client no longer skips the others. Notifications still fire from `submit_cut_bounds` after the cut is committed, not from `/approve`.

Saksham-Sirohi (approved):
The merge from `dev` keeps webhook delivery on `submit_cut_bounds` after the cut is committed, with per-client enqueue failures still isolated. It does not move the trigger back to `/approve`.

PR #237, opened 2026-09-13
feat(studio): integrate talk metadata into topbar, add user dropdown, and remove layout bloat

ViRUS-0-0:
## Closes #231 
- **Integrated Talk Metadata**: Replaced the redundant 120px broadcast banner with breadcrumb navigation (`Talks / [Event] / [Room] / [Talk Title]`) and a schedule pill badge (`Date, Time (Duration)`).
- **Eliminated Bloat & Sidebars**: Removed the duplicate left "Talk Metadata" panel and the "Generated Media Assets" section to maximize video workspace.
- **Zero Vertical Scrolling**: Scaled the video player panel with responsive flexbox (`object-fit: contain`, `overflow: hidden`) so the player and timeline scrubber always fit above the fold without scrolling.
- **Native User Dropdown**: Replaced horizontal login buttons with a native `<details>`/`<summary>` circular profile bubble containing user info, dark mode toggle, and logout.
- **Simplified Terminology**: Replaced internal "(Gate 1)" and "(Gate 2)" text with plain labels (`Pending Review`, `Preview Ready`).

## Pictures

Screenshot 2026-09-14 at 1 56 20 PM

---

Screenshot 2026-09-14 at 1 56 08 PM

---

Screenshot 2026-09-14 at 1 56 02 PM

## Verification

- All 583 automated tests passed (`uv run pytest --ignore=tests/test_smoke.py`).

## Summary by Sourcery

Streamline the Studio workspace around integrated talk metadata, compact media controls, and an accessible user menu.

New Features:
- Add event, room, title, and schedule metadata to the Studio top bar.
- Provide a video source selector with download support for final, preview, and raw media.
- Add an authenticated user dropdown with account information, theme switching, and logout controls.

Bug Fixes:
- Normalize timezone-aware talk schedules to UTC when displaying Studio metadata.
- Improve Studio authorization by loading event ownership with the talk record.

Enhancements:
- Streamline the Studio layout by removing redundant metadata and generated-asset panels and keeping the player and timeline visible without vertical scrolling.
- Simplify lifecycle status and approval terminology by removing internal gate labels.
- Restrict speaker-mode layout behavior to requests explicitly rendered as speaker mode.

Tests:
- Expand Studio coverage for breadcrumbs, schedule formatting, media source controls, user-menu accessibility, and speaker-mode rendering.
- Improve UI test fixture cleanup.

## Summary by CodeRabbit

- **New Features**
  - Added an authenticated user menu with account details, role information, theme controls, and logout.
[...]

PR #244, opened 2026-09-15
feat: event-scoped API keys UI, SSO user identity & idempotent schedule sync

### LINKED ISSUE #238: feat: event-scoped API keys UI, multi-role SSO identity & idempotent schedule sync
## Background & Architecture Intent

As established with the maintainers, **VEditor** operates as an independent, standalone media processing platform capable of integrating seamlessly with external platforms such as **Eventyay**.

The integration architecture centers around four core pillars:
1. **Event-Scoped API Keys**: External platforms integrate using API credentials scoped strictly to a specific event (`models.Client.event_ids`).
2. **Self-Service Key Management**: Organizers manage, generate, and revoke API keys for their events directly through the VEditor Web UI without requiring CLI/terminal access. Keys expose created and last-used timestamps.
3. **Core Personas**: The conference video production workflow centers on two distinct roles:
   - **Organizer**: Full event management, talk editing, pipeline control, and credential management.
   - **Speaker**: Scoped to their single talk (`scope_type="talk"`), watching their preview cut and submitting approval or change requests.
4. **Clean SSO Identity**: When users jump from an external platform via short-lived SSO JWT, VEditor displays their actual identity (name/email) rather than a generic "SSO (Organizer)" placeholder.

---

### 1. Web UI for Event-Scoped API Key Management
* **Backend Endpoints**:
  - `GET /events/{event_id}/api-keys`: List active API keys scoped to the event (masked, created date, last used).
  - `POST /events/{event_id}/api-keys`: Generate a new API key scoped to `event_id`. Returns the unhashed key once in the response.
  - `DELETE /events/{event_id}/api-keys/{client_id}`: Revoke an existing API key.
  - Gated strictly by `require_role("organizer")` (or event creator/admin).
* **Frontend Web UI**:
  - Add an API Keys modal on the Events page (`/studio/events`).
  - "Generate New API Key" action displaying the key with a one-click Copy to Clipboard button.
  - Table of active keys with creation timestamp, last used timestamp, and a Revoke action.

### 2. Role Hierarchy & Personas
* **Role Hierarchy**:
  - Update `ROLE_HIERARCHY`: `user (0) < speaker (1) < organizer (2) < admin (3)`.
  - Update `create_sso_token()` and `decode_sso_token()` to accept `role in ("organizer", "speaker")`.
  - Account signup defaults to `Organizer` so new organizers can immediately create and manage events.

### 3. SSO Identity & Header Presentation
* **Identity Claims in SSO JWT**:
  - Allow `create_sso_token()` to accept optional `email: str | None` and `display_name: str | None`.
  - Persist identity in the decrypted session cookie (`veditor_session`).
* **Topbar UI Update (`base.html.jinja`)**:
  - Update user topbar display: If `user.email` or `user.display_name` is present in an SSO session, display the user's name/email alongside their role badge `[Organizer]` or `[Speaker]` instead of hardcoded `SSO (Organizer)`.

### 4. Idempotent Talk Reconciliation & External ID Mapping
* **External Identifier Mapping**:
  - Maintain `source: str | None` and `external_id: str | None` on `models.Event` and `models.Talk`.
  - In `POST /talks/schedule/import`:
    - Deduplicate talks in-memory within the incoming payload.
    - Reconcile talks using `(event_id, external_id)`: update title, room, start, and end times if the talk already exists; insert if new.
    - Resolve target event by either internal integer ID or external slug.
  - Allow `/studio?event_id=<slug>` to resolve external slugs gracefully without 422 validation errors.

---

## Acceptance Criteria
- [x] Organizers can generate, view (one-time copy), list (with created/last-used timestamps), and revoke event-scoped API keys directly in the VEditor Web UI.
- [x] Role hierarchy aligns to `user (0) < speaker (1) < organizer (2) < admin (3)`.
- [x] SSO sessions display user identity (email/name) in the topbar when provided.
- [x] `POST /talks/schedule/import` reconciles talks idempotently via `external_id` without constraint conflicts.
- [x] Comprehensive unit, integration, and UI tests for API key management and schedule import.

SxxAq:
## Summary
Closes #238

Implements the unified external platform integration architecture agreed with maintainers:
1. **Event-Scoped API Key Management (UI & Backend)**:
   - Added UI modal on the Events page (`/studio/events`) allowing organizers to generate, list, regenerate, and revoke event-scoped API keys with one click.
   - Enforces a single active API key per event with confirmed regeneration and revocation dialogs.
   - Added `created_at` and `last_used_at` metadata timestamps to `models.Client`, schemas, and the UI table.
   - Clean, XSS-safe DOM table construction using `textContent` for dynamic values and status messages.
   - Added endpoints `GET /events/{event_id}/api-keys`, `POST /events/{event_id}/api-keys`, and `DELETE /events/{event_id}/api-keys/{client_id}`, guarded by `require_role("organizer")` and `check_event_access`.
   - Strictly restricted SSO sessions from generating or revoking permanent machine API keys.

2. **Role Hierarchy & Personas (`organizer`, `speaker`)**:
   - Explicit `ROLE_HIERARCHY`: `user (0) < speaker (1) < organizer (2) < admin (3)`.
   - Web signup defaults new accounts to `Organizer` so organizers can immediately create events and manage credentials.
   - SSO token creation and validation strictly supports `organizer` (event-scoped) and `speaker` (talk-scoped).
   - Removed `reviewer` role across schemas, models, routes, and templates.

3. **Rich SSO User Identity**:
   - SSO JWTs carry optional user identity claims (`email`, `display_name`).
   - Topbar dynamically displays the authenticated user's display name or email with their role badge instead of the static `"SSO (Organizer)"` string.
   - Preserved strict rejection of database injection claims (`user_id`, `sub`).

4. **External Event Mapping & Idempotent Schedule Reconciliation**:
   - Reconciles talks via `external_id` (e.g. Eventyay talk code) avoiding unique constraint collisions across imports.
   - Studio dashboard route gracefully handles string `event_id` slugs and external IDs.
   - Restricts ambiguous fallback imports to machine clients with exactly one event.

## Testing & Verification
- Unit test coverage for event API key listing, creation, regeneration, and revocation with metadata timestamps.
- Security tests for organizer and speaker SSO token generation, decoding, and identity claim propagation.
- Template lint and CSP compliance verified (zero inline styles or scripts in new UI elements).
- Full test suite passed (**679 unit/integration tests**).
- Ruff check and format pass without errors (`uv run pre-commit run --all-files`).


SxxAq:
https://github.com/user-attachments/assets/a319f5dd-6fbc-4185-8a34-d3e0e8e04b71

Saksham-Sirohi (changes requested):
### Review: Changes Requested :warning:

Great work implementing event-scoped API keys, XSS-safe key management in the UI, and idempotent schedule sync with external IDs.

There are a few critical issues around client deletion logic, performance, and cache headers that need to be addressed before merge:

#### 1. Data Loss: Unintended Deletion of Multi-Event API Clients (`app/routes/events.py`)
In `create_event_api_key`:
```python
for existing_c in all_clients:
    if event_id in (existing_c.event_ids or []):
        db.delete(existing_c)
```
And in `revoke_event_api_key`:
```python
db.delete(client)
```
If a client has access to multiple events (e.g. `client.event_ids = [1, 2]`), regenerating or revoking a key for event 1 will call `db.delete(client)`, **completely deleting the client record and inadvertently destroying access to event 2 as well**.
- **Fix**:
  - If `len(client.event_ids) > 1`: remove `event_id` from `client.event_ids` and commit.
  - Only execute `db.delete(client)` if `client.event_ids == [event_id]`.

#### 2. DB Write Contention in `get_client` Auth Dependency (`app/auth.py`)
In `get_client`:
```python
client.last_used_at = datetime.now(UTC)
try:
    db.commit()
except SQLAlchemyError:
    db.rollback()
```
`get_client` is executed on **every single authenticated request** that presents an API key. For high-frequency polling (e.g., job progress) or media chunking, executing an `UPDATE clients ...; COMMIT;` on every read request creates unnecessary write contention and database locks on the `clients` table.
- **Fix**: Throttle the update (e.g. only update `last_used_at` if it is `None` or older than 5 minutes).

#### 3. Security: Authenticated Waveform Cache Header (`app/routes/studio.py:545`)
In `app/routes/studio.py`:
```python
headers={"Cache-Control": "public, max-age=3600"}
```
Authenticated endpoints returning private media artifacts must not be marked `public` as intermediate shared proxies or CDNs may serve authorized waveforms to unauthorized users.
- **Fix**: Change to `private, max-age=3600` or `no-store`.

#### 4. Rebase and Merge Topbar User Dropdown (`base.html.jinja`)
PR #237 introduced an accessible `<details>`/`<summary>` profile bubble in the topbar. PR #244 replaces the topbar text with `user.email or user.display_name` in the old layout without the dropdown.
- **Action**: Please rebase onto #237 and integrate the dynamic `display_name` support into the new `<details>` user menu dropdown.

ViRUS-0-0 (changes requested):
Please resolve conflicts

app/routes/talks.py
```diff
@@ -1327,81 +1449,133 @@ async def import_schedule(
...
+    if talk_identifier.isdigit():
+        talk = (
+            db.query(models.Talk).filter(models.Talk.id == int(talk_identifier)).first()
+        )
+    if not talk:
+        talk = (
+            db.query(models.Talk)
+            .filter(models.Talk.external_id == talk_identifier)
+            .first()
+        )
```
ViRUS-0-0: The unique constraint on `Talk` is composite across `('event_id', 'external_id')` (`uq_talks_event_id_external_id`), meaning different events frequently have talks with identical external identifiers (e.g. 'TALK-01', '101', 'keynote'). In `get_talk`, `get_talk_jobs`, and `create_talk_sso_token`, the lookup executes `db.query(models.Talk).filter(models.Talk.external_id == talk_identifier).first()` without restricting the query to the caller's authorized event IDs (`client.event_ids`). If two events possess a talk with the same external_id: (1) A scoped client owning Event B will receive an arbitrary talk from Event A; when the subsequent authorization check `talk.event_id not in client.event_ids` runs, legitimate requests will be falsely rejected with 403 Forbidden; (2) For platform clients or admins, this causes cross-tenant leakage where the wrong talk or speaker preview is retrieved and modified.

app/routes/events.py
```diff
@@ -148,43 +165,197 @@ def delete_event(
...
+            db.query(models.Event)
+            .filter(models.Event.id == int(event_identifier))
+            .first()
+        )
+    if not event:
+        event = (
+            db.query(models.Event)
+            .filter(models.Event.external_id == event_identifier)
+            .first()
+        )
```
ViRUS-0-0: In `create_event_sso_token` (`app/routes/events.py`), the endpoint resolves the target event using `db.query(models.Event).filter(models.Event.external_id == event_identifier).first()`. Because the unique constraint is composite on `('source', 'external_id')` (`uq_events_source_external_id`), multiple external integrations (e.g. Pretix and Eventyay) can share the same slug. If a scoped client attempts to mint an SSO token for its event, and another event with the same slug exists, `first()` can return the foreign event, leading to an immediate 403 Forbidden rejection. For platform clients, this can cause minting an organizer token for the wrong event.

app/routes/events.py
```diff
@@ -148,43 +165,197 @@ def delete_event(
...
+    if user.is_sso:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="SSO sessions are not permitted to manage API keys",
+        )
+    check_event_access(event_id, user, db)
+    all_clients = (
+        db.query(models.Client).filter(models.Client.is_platform.is_(False)).all()
     )
+    event_clients = [c for c in all_clients if event_id in (c.event_ids or [])]
```
ViRUS-0-0: In both `list_event_api_keys` and `create_event_api_key`, the code executes `db.query(models.Client).filter(models.Client.is_platform.is_(False)).all()` and subsequently filters by `event_id in (c.event_ids or [])` in Python. This performs a full table scan, deserializing every non-platform client and hashed credential in the database into application memory on every key listing and generation request. As client registrations grow, this will cause memory pressure, connection latency, and degraded throughput.

app/auth.py
```diff
@@ -91,6 +100,26 @@ def get_client(
...
+        )
+        if (now - last_used).total_seconds() > 300:
+            should_update = True
+
+    if should_update:
+        client.last_used_at = now
+        try:
+            db.commit()
+        except SQLAlchemyError:
+            db.rollback()
```
ViRUS-0-0: Inside `get_client` (`app/auth.py`), `db.commit()` is called directly to update `client.last_used_at`. FastAPI dependency injection runs `get_client` prior to endpoint business logic. If a request handler expects an isolated transaction or encounters an error requiring rollback, the session was already prematurely committed. Furthermore, if `db.commit()` throws an exception, `db.rollback()` is executed, which expires all ORM instances on the session and leaves `client` detached. Accessing `client.event_ids` subsequently in route handlers can trigger `DetachedInstanceError` or database re-queries in an invalidated session.

app/routes/auth.py
```diff
@@ -176,6 +176,7 @@ def signup_submit(
     email: Annotated[str, Form()] = "",
     password: Annotated[str, Form()] = "",
     password_confirm: Annotated[str, Form()] = "",
+    role: Annotated[str, Form()] = "user",
```
ViRUS-0-0: According to the issue specification and PR description: 'Web signup defaults new accounts to Organizer so organizers can immediately create events and manage credentials.' While `signup.html.jinja` presents `<option value="organizer" selected>`, the backend route parameter in `app/routes/auth.py` defaults to `role: Annotated[str, Form()] = "user"`, and falls back to `"user"` if empty. Any programmatic signup or form submission omitting the role field results in an unprivileged account with `role="user"`.

app/ui/static/js/events.js
```diff
@@ -64,4 +64,229 @@ document.addEventListener('DOMContentLoaded', () => {
...
+          body: JSON.stringify({ name: 'Platform Integration Key' })
+        });
+        if (resp.ok) {
+          const data = await resp.json();
+          if (newKeyInput) newKeyInput.value = data.api_key;
+          if (newKeyAlert) newKeyAlert.classList.remove('api-key-alert-hidden');
+          loadApiKeys(currentActiveEventId);
+        } else {
+          alert('Failed to generate key');
+        }
```
ViRUS-0-0, line 265: In `app/ui/static/js/events.js`, `btnGenerateKey.addEventListener` invokes `loadApiKeys(currentActiveEventId)` without `await`. As a result, the `finally` block executes immediately while the network request is still pending. The button's disabled state is removed and `textContent` is evaluated before `loadApiKeys` updates the `data-has-key` attribute, leading to visual flickering and button text being incorrectly reset to '+ Generate API Key' instead of '↻ Regenerate Key'.

app/routes/events.py
```diff
@@ -148,43 +165,197 @@ def delete_event(
...
+    db.add(client)
+    db.commit()
+    db.refresh(client)
+
+    return schemas.ApiKeyCreatedResponse(
+        id=client.id,
+        name=client.name,
+        api_key=raw_api_key,
+        event_id=event_id,
+    )
```
ViRUS-0-0, line 336: The schema `ApiKeyCreatedResponse` defines an optional `created_at: datetime | None = None` field. However, `create_event_api_key` does not pass `created_at=client.created_at` when constructing the response model, resulting in `created_at` returning as `null` to API callers upon initial key generation.

app/ui/static/js/events.js
```diff
@@ -64,4 +64,229 @@ document.addEventListener('DOMContentLoaded', () => {
...
+          const keyId = b.getAttribute('data-key-id');
+          if (!confirm('Are you sure you want to revoke this API key? External integrations using this key will stop working.')) {
+            return;
+          }
+          b.disabled = true;
+          b.textContent = 'Revoking...';
+          try {
+            const delResp = await fetch(`/events/${eventId}/api-keys/${keyId}`, { method: 'DELETE' });
+            if (delResp.ok) {
+              loadApiKeys(eventId);
```
ViRUS-0-0: When an API key is revoked in `app/ui/static/js/events.js`, the alert `#new-key-alert` is not hidden and `#new-key-input` is not cleared. If a user generates a key and immediately revokes it, the revoked key remains visible with the 'New API Key Generated' success banner. Furthermore, closing the modal does not clear the sensitive key text from the input DOM node.

ViRUS-0-0 (changes requested):

app/routes/auth.py
```diff
@@ -230,6 +230,8 @@ def signup_submit(
 
     hashed = hash_password(password)
 
+    # All public registrations strictly receive the default 'user' role.
+    # Elevated roles (organizer, admin) must be granted by an administrator.
```
ViRUS-0-0, line 234: Unnecessary comment

app/cli.py
```diff
@@ -83,6 +83,30 @@ def create_client(
     print("Store this key safely! It will not be shown again.")
 
 
+def create_platform_client(session: Session, name: str):
```
ViRUS-0-0, line 86: Is creating a platform client still needed since the flow was changed? 

Each event gets its own api key so what's the use of platform client?

ViRUS-0-0 (changes requested):

app/cli.py
SxxAq: removed

ViRUS-0-0 (changes requested):
Resolve Conflicts Please

app/routes/events.py
```diff
@@ -148,43 +165,206 @@ def delete_event(
...
+    check_event_access(event_id, user, db)
+    event_clients = (
+        db.query(models.Client)
+        .filter(
+            models.Client.is_platform.is_(False),
+            models.Client.event_ids.any(event_id),
+        )
+        .all()
+    )
+    event_clients = [c for c in event_clients if event_id in (c.event_ids or [])]
```
ViRUS-0-0: Redundant in-memory list filtering in `list_event_api_keys` after SQL has already filtered by `event_id`.

Remove the second filtering pass `event_clients = [c for c in event_clients if event_id in (c.event_ids or [])]`. The database query already filtered results using `models.Client.event_ids.any(event_id)`.

app/routes/studio.py
```diff
@@ -315,6 +315,21 @@ def dashboard(
...
+            resolved_event_id = event_id
+        elif str(event_id).isdigit():
+            resolved_event_id = int(event_id)
+        else:
+            ev = (
+                db.query(models.Event)
+                .filter(models.Event.external_id == str(event_id))
+                .first()
+            )
+            resolved_event_id = ev.id if ev else -1
```
ViRUS-0-0: Dead code `isinstance(event_id, int)` in `dashboard` endpoint.

Remove `if isinstance(event_id, int):`. In FastAPI, the route parameter `event_id` is declared as `str | None = None`. Query parameters parsed from URLs are always `str` or `None`, so `isinstance(event_id, int)` can never evaluate to `True`.

app/routes/talks.py
```diff
@@ -1320,41 +1370,122 @@ async def import_schedule(
...
+                if existing_event:
+                    if not is_platform:
+                        check_event_access(existing_event.id, user, db)
+                    event = existing_event
+                else:
+                    created_by = user.user_id if not user.is_machine else None
+                    event = models.Event(name=event_name, created_by_user_id=created_by)
+                    db.add(event)
+                    db.flush()
+                    is_new_event = True
```
ViRUS-0-0: Redundant `target_str.isdigit()` check and duplicate event creation logic in `import_schedule`.

Remove `if target_str.isdigit():` and its duplicate `else` creation block. Earlier in the function, `is_explicit_external` is set to `True` whenever `target_identifier` is not digits. In this `else` branch, `target_str.isdigit()` is already guaranteed to be `True`. Furthermore, creating an event inside that nested `else` block duplicates the fallback event creation at lines 1524–1530 and bypasses name-based matching.

app/routes/talks.py
```diff
@@ -1320,41 +1370,122 @@ async def import_schedule(
...
 
     if not event:
         # Check if caller already has an event with matching name
-        if user.is_machine:
+        if getattr(user, "is_platform", False):
+            event = (
+                db.query(models.Event).filter(models.Event.name == event_name).first()
+            )
+        elif user.is_machine:
```
ViRUS-0-0: Duplicate database query branches for platform clients and admin users in `import_schedule`.

 Merge `getattr(user, "is_platform", False)` and `user.role == "admin"` into a single branch `if getattr(user, "is_platform", False) or user.role == "admin":`.

app/routes/talks.py
```diff
@@ -145,8 +145,23 @@ def get_talk(
...
+            user.is_machine
+            and not user.is_platform
+            and user.event_ids
+            or not user.is_machine
+            and not user.is_platform
+            and not user.is_human_admin
+            and user.event_ids
+        ):
+            talk_filters.append(models.Talk.event_id.in_(user.event_ids))
+        talk = db.query(models.Talk).filter(*talk_filters).first()
```
ViRUS-0-0: Duplicated talk lookup logic and over-complicated boolean expression in `get_talk` and `get_talk_jobs`.

Simplify the boolean expression from `(user.is_machine and not user.is_platform and user.event_ids or not user.is_machine and not user.is_platform and not user.is_human_admin and user.event_ids)` to `(not user.is_platform and not user.is_human_admin and user.event_ids)`. Extract the lookup into a single helper rather than duplicating it across endpoints.

Saksham-Sirohi (approved):

app/routes/talks.py
```diff
@@ -1399,81 +1543,150 @@ async def import_schedule(
...
     if not talk:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
             detail="Talk not found",
         )
-    if talk.event_id not in (client.event_ids or []):
+    if not getattr(client, "is_platform", False) and talk.event_id not in (
+        client.event_ids or []
+    ):
```
ViRUS-0-0: Re-implementing existing `verify_event_access` helper inline in `create_talk_sso_token` and `create_event_sso_token`.

Replace the manual `getattr` and list-membership checks with a direct call to [`verify_event_access(talk.event_id, client)`] from [`app/auth.py`]

PR #245, opened 2026-09-15
fix: _parse_hhmmss() silently parses 24:00:00 as 0.0 instead of rejecting invalid hours 

### LINKED ISSUE #243: _parse_hhmmss() silently parses 24:00:00 as 0.0 instead of rejecting invalid hours
## Description

When submitting cut bounds, passing an hour value of `24` (e.g., `24:00:00`) for `cut_start` or `cut_end` is accepted by the API instead of returning a validation error.

`_parse_hhmmss()` relies on `datetime.time.fromisoformat()`, which interprets hour `24` as `0`. As a result, a cut intended to begin 24 hours into a long recording is silently persisted as `cut_start = 0.0` (the beginning of the file), producing incorrect video clips without alerting the user.

## Location

* **Backend Module:** Review Studio / Talk Cut API endpoint
* **Function:** `_parse_hhmmss()`
* **API Endpoint:** `POST /talks/{id}/cut`

## Steps to Reproduce

1. Select a talk in `pending_bounds` state.
2. Send a POST request to `/talks/{id}/cut` with an hour value set to `24`.
3. Observe that the API returns a `202 Accepted` response and persists `cut_start` as `0.0`.

## Current Behavior

* `_parse_hhmmss("24:00:00")` evaluates to `0.0` via `time.fromisoformat()`.
* The API accepts the payload (`202 Accepted`) and overwrites the start time with `0.0`.

## Expected Behavior

* Hours greater than or equal to 24 (`HH >= 24`) or invalid `HH:MM:SS` strings should fail validation.
* The API should reject the payload with a `422 Unprocessable Entity` (or `400 Bad Request`) error.

## Proposed Fix

Update `_parse_hhmmss()` to explicitly validate that hour values fall within the expected range (`00-23` for standard clock time, or custom duration parsing if total seconds spanning multiple days are supported):

```python
def _parse_hhmmss(time_str: str) -> float:
    parts = time_str.split(":")
    if len(parts) != 3:
        raise ValueError("Invalid time format. Expected HH:MM:SS")

    hours, minutes, seconds = map(float, parts)
    if hours >= 24 or minutes >= 60 or seconds >= 60:
        raise ValueError("Time components out of range (HH must be < 24)")

    return hours * 3600 + minutes * seconds

```

## Acceptance Criteria

* [ ] Add explicit boundary checks in `_parse_hhmmss()` for `HH >= 24`, `MM >= 60`, and `SS >= 60`.
* [ ] Return a `422 Unprocessable Entity` response when invalid time strings like `24:00:00` are sent to `POST /talks/{id}/cut`.
* [ ] Add unit tests verifying `_parse_hhmmss()` raises validation errors for out-of-range time strings.

arc12at:
## Closes #243 

This PR fixes silent coercion of HH >= 24 in _parse_hhmmss() to 0.0. Out-of-range times submitted to POST /talks/{id}/cut now return 422 Unprocessable Entity instead of persisting cut_start = 0.0.

## Summary by Sourcery

Validate cut times strictly and prevent invalid or precision-collapsed values from being persisted.

Bug Fixes:
- Reject out-of-range and malformed HH:MM:SS values instead of silently coercing invalid hours such as 24:00:00 to zero seconds.
- Return HTTP 422 for invalid cut bounds submitted to the talk-cut endpoint.

Enhancements:
- Preserve high-precision time ordering during validation and reject values whose ordering is lost when converted to floats.

Tests:
- Add schema and endpoint regression coverage for valid times, malformed formats, out-of-range components, strict parsing, and high-precision ordering.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for cut time inputs, including hours, minutes, seconds, and high-precision fractional seconds.
  * Malformed, whitespace-padded, newline-containing, or out-of-range values now return validation errors instead of being silently accepted or coerced.
  * Cut ranges must have an end time later than the start time, including after numeric conversion.
  * Values too close to preserve their ordering are now rejected.

* **Tests**
  * Added coverage for valid parsing, boundary values, malformed inputs, precision limits, and cut request validation.


ArnavBallinCode (approved):
LGTM

ViRUS-0-0 (approved):
LGTM

PR #260, opened 2026-09-16
feat(admin): runtime settings management

### LINKED ISSUE #252: [ADMIN] Runtime Settings Management
**Description:**
Allow administrators to tweak global platform settings (like default preview presets, video tolerances) dynamically from the UI without requiring a server restart or manual edits to environment variables.

**Tasks:**
1. Create a new SQLAlchemy model `SystemSetting` (e.g., `key`, `value`, `description`, `updated_at`) in `app/models.py`.
2. Generate an Alembic migration (`alembic revision --autogenerate -m "add system settings"`) and apply it.
3. Update configuration access patterns (e.g., in `app/config.py` or via a DB helper) so that the application checks the database for a setting override before falling back to environment variables.
4. Build the `/admin/settings` UI form to list, edit, and save these settings securely.

**Scope:**
- Global configuration overrides that are safe to change at runtime.

**Outcome:**
Admins can adjust platform behavior instantly, making it easier to adapt to unexpected event requirements (e.g., tweaking audio tolerance if a recording is unusually quiet, or changing default video encoders).

**Out of scope:**
- Event-specific settings overrides (these belong directly on the `Event` model, like `retention_overrides`).
- Secrets management (passwords, JWT secrets, and API keys must remain in environment variables).

**Acceptance Criteria:**
- The database schema is successfully created and migrated.
- Admins can view a list of editable settings in the UI.
- Form submissions successfully update the database.
- The video processing pipeline respects the new settings immediately on subsequent job runs.

**Dependency on any other issue:**
- Depends on: **Sub-issue 1** (Admin Scaffolding)

sanketkokate712:
## Closes #252

Admins can now change a set of platform settings from `/admin/settings`. There's no server restart and no change to environment variables. Pipeline workers use the new value starting with the next job.

### What changed

**Data model**
- New `SystemSetting` model (`key` PK, `value` JSONB, `description`, `updated_at`, `updated_by_user_id` → `users.id` `ON DELETE SET NULL`).
- Migration `0011_add_system_settings`. I tested upgrade, downgrade, and upgrade again on a fresh DB, and `alembic check` reports no drift between the model and the migration.

**Settings lookup: `app/runtime_settings.py`**
- Only settings in an **allowlist** (`RUNTIME_SETTINGS`) can be overridden. Each entry has a type, bounds or allowed choices, a label and a description. Secrets such as `SESSION_SECRET`, DB/Redis URLs and JWT settings aren't in the list, so they can't be changed from the UI (the form route returns 404 for them).
- `get_setting(key)` returns the **DB override** if a valid one exists. Otherwise it returns the **env/code default**, which is read at call time, so `.env` values still apply.
- **No cache.** Each lookup is one primary-key read per job run, so a change applies to the next job without cache invalidation across workers.
- **Safe fallbacks:** if a stored value is invalid, or the DB read fails (for example, the migration hasn't run yet), it logs a warning and uses the default. The lookup opens its own short-lived session, so a failed read can't break the caller's transaction.

| Setting | Type / range | Default | Used by |
|---|---|---|---|
| `detect_duration_tolerance_seconds` | 0 – 86400 s | 300 | `job_detect` |
| `loudness_target_lufs` | -70 – 0 LUFS | -16 | `job_loudness` |
| `preview_preset` | `settings.preview_presets` keys | `small_video` | `job_preview` |
| `transcode_preset` | `1080p_default` / `720p` / `4k_master` | `1080p_default` | `job_transcode` |
| `disk_guard_multiplier` | 1 – 100 | `DISK_GUARD_MULTIPLIER` env (3.0) | ingest disk guard |

To support `transcode_preset`, I added a `TRANSCODE_PRESETS` name → preset map next to the existing preset constants.

**Admin UI / routes (`app/routes/admin.py`, router-level `require_admin`)**
- `GET /admin/settings` shows one card per setting: the effective value, an **Override** or **Default** badge, the default value, the allowed range, and when the value was last changed and by whom.
- `POST /admin/settings/{key}`:
  - `action=save` validates the value on the server. An invalid value returns 400 with the error shown on the page, and nothing is saved. A valid value is saved with an `INSERT … ON CONFLICT DO UPDATE`, so two admins saving the same key at the same time can't hit a primary-key error. Then it redirects (303) to the page with a confirmation message.
  - `action=reset` deletes the override, so the setting goes back to its default.
- Security: only a human admin can use these routes. API keys and SSO sessions get 403, and anonymous requests get 401. Session cookies are `SameSite=Lax`, so another site can't submit the form with an admin's cookie. The HTML inputs also have `min`/`max`, but the server check is what enforces the limits.
- Added a "Settings" sidebar link, shown to admins only.

### Tests
- New `tests/test_admin_settings.py` (27 tests):
  - access control (401/403)
  - page rendering
  - save, repeated save (upsert), and who made the change
  - invalid values: out of range, NaN, not a number, empty, unknown choice
  - unknown key or action
  - reset
  - fallback to the env default, to the default for a bad stored value, and to the default on a DB error
  - **the pipeline uses the override on its next run**, for detect, preview, loudness, transcode and the ingest disk guard
- `tests/test_tasks.py`: updated the stub pipeline functions to accept the new keyword arguments.
- Full suite: **749 passed**. `test_e2e_heavy_job_in_flight_does_not_block_light_job` and `test_smoke.py` also fail on unmodified `dev` because they need a live server on localhost:8000, so this change didn't cause them.
- `ruff check` and `ruff format` pass.

### Screenshots (local Docker stack)

**Default state**
defaults

**Server-side validation (loudness above 0 LUFS is rejected, nothing saved)**
validation

**Overrides saved (with who changed each one and when)**
saved

**Reset back to default**
reset

**Dark mode**
dark

**Narrow viewport**

I also checked the running container: `get_setting()` returned the values saved in the UI (`loudness_target_lufs = -14.0`, `transcode_preset = 720p`) and the default for the setting I reset.

### Out of scope (per issue)
- Per-event overrides (these stay on `Event`, e.g. `retention_overrides`).
- Secrets management.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Enable human administrators to manage allowlisted runtime platform settings from the admin UI without server restarts.

New Features:
- Add an admin interface for viewing, overriding, and resetting an allowlisted set of runtime platform settings without restarting the server.
- Apply runtime overrides to detection, preview generation, loudness normalization, transcoding, and ingest disk checks on subsequent jobs or requests.

Enhancements:
- Resolve settings safely from database overrides with validation, default fallbacks, per-call lookup, and audit metadata for the administrator who made each change.
- Restrict runtime settings management to human administrators and provide server-side validation with concurrency-safe updates.

Tests:
- Add coverage for settings access control, validation, persistence, reset behavior, fallback handling, audit information, and pipeline adoption of overrides.

Chores:
- Add the system settings data model and database migration.

## Summary by CodeRabbit

* **New Features**
  * Added an admin Settings page for managing runtime processing options.
  * Administrators can save, validate, and reset settings, with current values, defaults, and update details displayed.
  * Processing jobs now use configurable detection, preview, loudness, transcoding, and disk-space parameters.
  * Added support for predefined transcoding presets, including 1080p, 720p, and 4K options.
* **Bug Fixes**
  * Invalid or unavailable saved settings now safely fall back to defaults.


ViRUS-0-0:
Thanks for the contribution. This is not the implementation we want. Closing this PR, as this issue is not open for new contributors. 

Also, I see you have raised about 5-6 prs in eventyay as well. Please raise upto three PRs here and there as well.

sanketkokate712:
> Thanks for the contribution. This is not the implementation we want. Closing this PR, as this issue is not open for new contributors.
> 
> Also, I see you have raised about 5-6 prs in eventyay as well. Please raise upto three PRs here and there as well.

ok will be implementing this rule onwards, thanks!

PR #240, opened 2026-09-14
fix: enforce safe login redirects and consistent studio route authentication

arc12at:
## Closes #229 

This PR resolves security and navigation issues across VEditor Studio routes (/studio, /studio/events, and /studio/talks/{id}).

## Acceptance Criteria

- [x] Visiting /studio or /studio/events in an incognito window immediately redirects to /login with the respective next parameter.
- [x] Submitting valid login credentials lands the user directly on the page they originally requested.
- [x] No talk data, event statistics, or room information is returned to unauthenticated callers.
- [x] External URLs in next are ignored, preventing open-redirect vulnerabilities.
- [x] Automated tests pass:
`uv run pytest tests/test_routes_studio.py tests/test_auth.py
`

## Videos/photo

https://github.com/user-attachments/assets/492f161f-e1d7-46bb-9f8e-a03cfc92ce63

---

https://github.com/user-attachments/assets/87ad0665-378c-4b38-a3cc-ef0baa1bfa1b

---

image

## Summary by Sourcery

Secure Studio route access and restore users to their requested destination after successful login.

New Features:
- Preserve validated internal destination URLs through login so users return to the Studio page they originally requested.

Bug Fixes:
- Prevent unauthenticated callers from accessing or receiving Studio talks, events, statistics, and room data.
- Prevent open redirects by rejecting external and protocol-relative login destinations.
- Handle invalid API keys and unauthorized SSO sessions with explicit authentication or authorization responses.

Enhancements:
- Align browser, API-key, cookie, and SSO access behavior across Studio dashboard, event management, and talk detail routes.
- Show an in-dashboard authorization error when users without organizer privileges attempt to access event management.

Tests:
- Add coverage for Studio authentication redirects, data-leak prevention, API-key and SSO authorization, role restrictions, and login destination validation.

## Summary by CodeRabbit

* **New Features**
  * Sign-in now returns users to their originally requested page, including after failed attempts.
  * Studio supports authenticated API-key and SSO access across supported features.
  * Dashboard displays clear error alerts for unauthorized actions.

* **Bug Fixes**
  * Unauthenticated Studio requests redirect to sign-in with their destination.
  * Invalid API-key cookies redirect to sign-in instead of returning an authentication error.
  * Invalid or empty SSO tokens receive clear authentication errors.
  * Unsafe external redirects fall back to Studio.
  * Unauthorized event access now redirects with an explanatory message.

* **Tests**
  * Expanded coverage for sign-in redirects and Studio authentication.

PR #253, opened 2026-09-16
refactor: speaker role, pipeline handoff and review flow

ViRUS-0-0:
## Closes #246 

> [!WARNING]
> -  This is a stacked pr and will be ready for review after #241 is merged

https://github.com/user-attachments/assets/59ba7b20-55d2-4e6b-bdc0-2648acf97802

---

https://github.com/user-attachments/assets/61e0f668-bfb0-4bed-83e7-97be51a5aba1

## Summary by Sourcery

Establish speaker-specific talk ownership and access while separating organizer handoff from the review and assembly workflow.

New Features:
- Add speaker identities and talk assignments, including speaker-specific authentication, access control, dashboard filtering, and SSO support.
- Introduce an organizer-to-speaker handoff step for configuring intro/outro branding and assigning a speaker before timestamp review.
- Add speaker-focused studio UI, preview-quality guidance, review handoff controls, abort/reset confirmations, and restricted management actions.

Bug Fixes:
- Automatically start final assembly after preview approval and record dispatch failures as broken jobs with persisted logs.
- Allow talks needing changes to resubmit cut bounds and reset rejected previews to the raw-video review stage.

Enhancements:
- Restructure the talk lifecycle to separate organizer handoff, speaker timestamp submission, preview review, and final assembly.
- Restrict speaker access to assigned talks and hide sensitive job log paths while preserving status and progress visibility.
- Normalize speaker email assignments and support importing them from schedule data.

Tests:
- Update lifecycle, authorization, security, model, and studio UI tests for speaker assignments, handoff flow, access restrictions, and assembly dispatch.

Chores:
- Add a database migration for the speaker role and talk speaker email field.

## Summary by CodeRabbit

- **New Features**
  - Added speaker roles and speaker email assignment for talks.
  - Speakers can access assigned talks, review content, submit cut boundaries, and complete handoff steps.
  - Added speaker-specific dashboard and studio views, including preview quality notices.
  - Added optional speaker email during talk creation and import.

- **Workflow Updates**
  - Approvals now proceed through intro/outro configuration before production.
  - Rejected or reset talks return to editable boundary stages.
  - Preview, raw, and final video cut handling is more accurate.
  - Assembly failures are recorded for follow-up.

- **UI Updates**
  - Added handoff, abort, and reset-to-raw actions.
  - Management controls are hidden from non-manager users.
  - Job log paths are hidden from speakers.

PR #262, opened 2026-09-16
feat(studio): link event and room names to filtered talk pages

### LINKED ISSUE #256: Add Room and Event Page in Talk studio navigation bar
## Description

Currently, in the Talks overview/dashboard, the **Event Name** and **Room Name** associated with each talk are displayed as static text. To improve navigation and organization for users managing multiple tracks and events, these labels should be converted into clickable links.

Clicking the Event Name should navigate the user to the specific Event's page, and clicking the Room Name should navigate to a newly created Room page. Both of these pages will act as filtered views, displaying a title header and a list of all talks associated with that specific Event or Room.

## Current Behaviour

- Event Name and Room Name text on the Talks dashboard are plain, non-interactive text elements.
- There is no dedicated isolated page to view talks filtered solely by a specific "Room".
- Users must manually search or visually scan the main dashboard to find other talks happening in the same room or event.

## Expected Behaviour

- **Event Name Links**: Clicking an Event Name on any talk card/row navigates the user to the respective event page, displaying the Event Title at the top and a filtered list of its talks below.
- **Room Name Links**: Clicking a Room Name navigates the user to a new route (e.g., `/studio/rooms/{room_name}` or a parameterized URL), displaying the Room Name at the top and a filtered list of talks scheduled for that room.

## Tasks

- [ ] Update the Jinja2 template for the Talks dashboard to wrap the Event Name and Room Name text in `<a>` tags with appropriate routing hrefs.
- [ ] Ensure the Event Name link navigates to the existing Event detail/talks page.
- [ ] Create a new FastAPI route in `app/routes/studio.py` to handle the new Room view.
- [ ] Create a new Jinja2 template for the Room page (or adapt an existing filtered dashboard template) that displays the Room title and its associated talks.
- [ ] Write tests to verify that the new Room route correctly filters and returns only talks for the specified room.
- [ ] Write UI tests to ensure clicking the new links correctly navigates to the expected filtered views.

## Acceptance Criteria

- [ ] Event Names are clickable and successfully route the user to a page showing only that event's talks.
- [ ] Room Names are clickable and successfully route the user to a newly implemented Room page.
- [ ] The new Room page displays the room name clearly in the header.
- [ ] The new Room page displays only the talks scheduled within that specific room.
- [ ] Styling for the new clickable links matches the VEditor design system (no inline CSS).
- [ ] All new routing and filtering logic is covered by unit/integration tests and passes CI.

## Related Picture

Image

sanketkokate712:
## Closes #256

Makes the event and room names in the Talk studio navigation bar clickable. The event name opens that event's talk list. The room name opens a new Room page that lists only the talks in that room.

### What changed

**Links**
- **Studio topbar breadcrumb** (`studio.html.jinja`): the event links to `/studio?event_id={id}` and the room links to `/studio/rooms/{room}?event_id={id}`. A talk with no room still shows `—`.
- **Dashboard rows** (`dashboard.html.jinja`): the room column is a link, and the `Event #id` label under each title now shows the event name as a link. Events are loaded with `selectinload`, so this adds no N+1 queries. Links only render when the talk has an `event_id`.
- For SSO sessions scoped to a single talk, the breadcrumb stays plain text. That session can't open event or room lists, so those links would only redirect back to the talk.

**Room page: `GET /studio/rooms/{room_name:path}`** (`app/routes/studio.py`)
- The dashboard's access-scoping code, including the auth gate from #240, moved into `_render_talks_page()`, and both `/studio` and the new route call it. The Room page therefore follows the same rules as the dashboard, and no authorization logic was copied:
  - Logged-out requests redirect to `/login?next=<current page>`, which is built from the path and query string and percent-encoded once. The user comes back to the same room, or the same filtered dashboard, after signing in.
  - An invalid `X-API-Key` gets a 401.
  - Organizers and admins see their own events, API-key clients see their assigned events, and event-scoped SSO sessions see their one event. Talk-scoped SSO sessions are redirected to their talk.
  - Anonymous requests never reach the talk query or the stats.
- The Room page uses the same template as the dashboard, with a **Room** heading and breadcrumb (`← Talks / Event / Room`). The status filter and search still work, and the filter form and the Clear button stay on the Room page and keep `event_id`.
- Links include `event_id`, so a room name used by two events (e.g. "Main Auditorium") only shows the current event's talks. Without `event_id`, the page lists that room's talks from the events the caller can access, and never from anyone else's.
- `{room_name:path}` allows room names that contain `/`.

**Event page (`/studio?event_id=`)**
- It now shows an **Event** heading with the event name and a breadcrumb. The name only appears for events the caller can access.
- **Behavior change:** on event and room pages, the stat cards now count only the talks in that view, and the Attach Room Video list only offers the selected event's rooms. Before, `/studio?event_id=` showed totals and rooms for the whole workspace above a filtered list. The unfiltered `/studio` dashboard is unchanged.
- Stats come from a `GROUP BY status` count and the room list from `SELECT DISTINCT room`, both scoped in SQL, so no talk rows are loaded to compute them.

**Styling:** there's no inline CSS. Breadcrumb link styles moved from `studio.css` to `app.css` because the dashboard doesn't load `studio.css`. `dashboard.css` gets `.talk-meta-link`, `.table-kicker` and `.table-title-scoped`. At rest, links have a dotted underline so they look clickable, and they turn primary blue on hover or focus. Both themes use existing tokens.

### Tests
There are 13 new tests in `tests/test_ui_events.py`:
- the Room page lists only that room's talks and excludes other rooms, talks with no room, and the same room name in another event
- the Room page is scoped to the caller: another organizer's event returns 0 results, and nothing leaks when `event_id` is omitted
- the status filter and search work on the Room page, and Clear keeps the room and event scope
- the Event page shows the event heading, and its stats count only that event (checked with an organizer who owns two events)
- **auth:** a logged-out request to the Room page redirects to login with the room URL as `next`, with or without `event_id`, and returns no talk data
- **auth:** an invalid API key gets a 401
- **auth:** logging in with the `next` taken from the actual redirect returns the user to the Room page, and to a filtered dashboard with `event_id` and `status_filter` intact
- **auth:** a talk-scoped SSO session opening a Room page is redirected to its own talk
- the Attach Room Video room list is scoped to the selected event
- **navigation:** the room and event links are read from the rendered dashboard HTML and followed. This includes a room named `Room B/2`.
- **navigation:** the event and room links in the studio breadcrumb are followed, and a talk with no room shows no room link
- talk-scoped SSO sessions get a plain-text breadcrumb

Results:
- Rebased on `dev`, including #240. Related suites (`test_ui_events`, `test_routes_studio`, `test_studio`, `test_sso`, `test_auth`, `test_auth_routes`, `test_events`, `test_admin_routes`): **228 passed**
- Full suite: **749 passed, 7 failed** (`test_smoke.py` was deselected because it needs a live server). None of the failures come from this PR:
  - Six are in `test_ingest.py` (symlinks, path separators) and `test_security.py` (file mode `0o600`). They're Windows-only and fail the same way on unmodified `dev`.
  - `test_e2e_heavy_job_in_flight_does_not_block_light_job` is a Redis worker-timing test that also fails locally with this PR's changes removed. It passes in CI.
- `ruff check` and `ruff format --check` pass.

### Screenshots
These were taken on a local server with seeded data, signed in as an organizer. There are two events, and both have a room called "Main Auditorium".

**1. Studio topbar:** the event and room are links (hovering the room)
studio topbar links

**2. Clicking the room opens `/studio/rooms/Main%20Auditorium?event_id=…`:** it has the room heading and 3 talks. The other event's Main Auditorium talk isn't listed.
room page

**3. Clicking the event in the breadcrumb opens the event page:** the event heading and stats cover only this event
event page

**4. Dashboard:** event and room are links in every row (hovering a room)
dashboard row links

**5. Room page, dark theme**
room page dark

## Summary by Sourcery

Enable event- and room-focused navigation throughout Studio while enforcing event visibility and preserving filtered views.

New Features:
- Add clickable event and room navigation links from the studio dashboard and talk breadcrumbs.
- Add room-scoped talk pages with event-aware filtering, search, status filters, and room headings.

Bug Fixes:
- Prevent unauthorized users and scoped clients from viewing talks, statistics, or rooms outside their permitted events.
- Preserve room names and filtered destinations correctly across URL navigation and login redirects.

Enhancements:
- Unify dashboard and room-page authorization and rendering behavior.
- Scope event-page statistics and room-video options to the selected event and compute them from filtered database queries.
- Add scoped-page breadcrumb, heading, and link styling for light and dark themes.

Tests:
- Add comprehensive coverage for room and event filtering, authorization, navigation, URL delimiter handling, login redirects, and SSO behavior.

## Summary by CodeRabbit

* **New Features**
  * Added room-specific talk views with contextual breadcrumbs and scoped titles.
  * Added room and event navigation links with safe handling of spaces and special characters.
  * Preserved filters and search criteria when navigating or signing in.
  * Added scoped room lists and status counts based on viewer access.

* **Bug Fixes**
  * Corrected redirects and filter links for special characters in room names.
  * Improved event and room breadcrumb behavior across dashboard and studio views.
  * Improved event matching within the viewer’s available events.


app/routes/studio.py
```diff
@@ -404,6 +449,12 @@ def dashboard(
         all_talks = db.query(models.Talk).all()
```
Saksham-Sirohi: Unauthenticated all_talks = Talk.query.all() leaks global stats/room enumeration; should be empty or unreachable after auth gate.

Saksham-Sirohi (changes requested):
Feature wiring looks correct: breadcrumbs/dashboard link to /studio?event_id= and /studio/rooms/{room}?event_id=; room_talks reuses _render_talks_page with room filter (studio.py:310-328,414-415). Talk-scoped SSO correctly keeps non-clickable breadcrumbs (studio.html.jinja). Blocking: unauthenticated branch still serves data. For no user/client, talks query only filters by event_id when provided, then by room — so GET /studio/rooms/{room} without auth/event_id lists matching rooms across all events; all_talks = db.query(Talk).all() then room-filtered (studio.py:410-415,448-456). Depends on #240 auth redirects which are not in this branch; new public room route expands disclosure surface.

app/routes/studio.py
sanketkokate712: Thanks for catching this. I've rebased on dev with #240 and fixed it in 2c3ee84.

The fix: the Room page and /studio now both use _render_talks_page(), and #240's auth gate is inside that function. So a logged-out request to /studio/rooms/… (with or without event_id) never runs the talk query or all_talks.

Logged out: it redirects to /login?next=<room URL>, and the user comes back to the same room after signing in. next goes through _get_safe_redirect_target.
Invalid X-API-Key: returns 401.
Logged in: the room filter only applies on top of the caller's own scope (own events for organizers and admins, assigned events for API-key clients, the one event for event-scoped SSO). Leaving out event_id can't list the room from other people's events.
The fallback all_talks = db.query(Talk).all() you flagged is gone. Anonymous requests return before that point, and #240 had already changed the fallback to an empty list.

New tests in tests/test_ui_events.py:

test_room_page_requires_authentication: redirect with and without event_id, and no talk data in the response
test_room_page_rejects_invalid_api_key: 401
test_login_next_returns_to_room_page: logging in returns to the Room page

SxxAq (changes requested):
#### 1. Push Event & Room Filters into SQL for `all_talks` (`app/routes/studio.py:440-471`)

**Problem:**
When calculating status metrics (`status_counts`) for an event or room page, `all_talks` queries the database for **all talks across all of the organizer's events** and then filters down in Python memory:
```python
if user.role in ("organizer", "admin"):
    org_event_ids = [e.id for e in user_events]
    all_talks = (
        db.query(models.Talk)
        .filter(models.Talk.event_id.in_(org_event_ids))
        .all()
    )
# ...
if event_id is not None:
    all_talks = [t for t in all_talks if t.event_id == event_id]
if room is not None:
    all_talks = [t for t in all_talks if t.room == room]
```
For accounts with multiple large events, loading thousands of talk rows into memory to count statuses for a single room is inefficient.

**Suggested Fix:**
Apply `event_id` and `room` filters directly in SQL when building the `all_talks` query:
```python
stats_query = db.query(models.Talk)
# Scope by organizer/admin/client
if sso_user:
    stats_query = stats_query.filter(models.Talk.event_id == sso_user["scope_id"])
elif user and user.role in ("organizer", "admin"):
    org_event_ids = [e.id for e in user_events]
    stats_query = stats_query.filter(models.Talk.event_id.in_(org_event_ids))
elif client:
    stats_query = stats_query.filter(models.Talk.event_id.in_(client.event_ids or []))
else:
    stats_query = stats_query.filter(models.Talk.id == -1)

# Push event and room filters to DB
if event_id is not None:
    stats_query = stats_query.filter(models.Talk.event_id == event_id)
if room is not None:
    stats_query = stats_query.filter(models.Talk.room == room)

scoped_talks = stats_query.all()
```

---

#### 2. Avoid Double URL-Encoding on Login Redirect (`app/routes/studio.py:331-332, 361`)

**Problem:**
In `room_talks`:
```python
login_next=urllib.parse.quote(str(request.url.path), safe="/")
+ (f"?{request.url.query}" if request.url.query else "")
```
And then in `_render_talks_page`:
```python
url=f"/login?next={urllib.parse.quote(login_next, safe='/')}"
```
Because `request.url.path` is already quoted in `room_talks`, calling `urllib.parse.quote()` a second time in `_render_talks_page` double-encodes spaces and special characters (`%20` becomes `%2520`, `?` becomes `%3F`).

**Suggested Fix:**
Pass the unquoted path into `login_next` in `room_talks`:
```python
login_next = str(request.url.path) + (f"?{request.url.query}" if request.url.query else "")
```
and let `_render_talks_page:361` perform the single `urllib.parse.quote(login_next, safe="/")` call.
```

File: app/routes/studio.py [hunk 2]
```diff
@@ -300,7 +300,52 @@ def dashboard(
...
+    return _render_talks_page(
+        request,
+        db,
+        client,
+        event_id=event_id,
+        status_filter=status_filter,
+        q=q,
+        room=room_name,
+        login_next=urllib.parse.quote(str(request.url.path), safe="/")
+        + (f"?{request.url.query}" if request.url.query else ""),
```
ViRUS-0-0: Do not pre-encode `request.url.path` in `room_talks`. Pass the raw path and query string to `_render_talks_page` and perform percent-encoding exactly once when constructing the `/login?next=...` redirect URL.

app/routes/studio.py
```diff
@@ -300,7 +300,52 @@ def dashboard(
...
+    status_filter: str | None,
+    q: str | None,
+    room: str | None = None,
+    login_next: str = "/studio",
+):
+    """Render the talks list, scoped to the caller and optionally to an event/room."""
+    # Check for authenticated user or active SSO session in cookie
     user = _get_authenticated_user_from_cookie(request, db)
     cookie_token = request.cookies.get("veditor_session")
     sso_user = decode_sso_token(cookie_token) if (not user and cookie_token) else None
```
ViRUS-0-0, line 412: In the existing dashboard route, SSO sessions scoped to an individual talk (`scope_type == 'talk'`) are intercepted and redirected directly to their assigned talk studio page (`/studio/talks/{talk_id}`). Because this auth check precedes `return _render_talks_page(...)` in `dashboard()` and was not migrated inside `_render_talks_page()`, a speaker holding a talk-scoped SSO token can navigate directly to `/studio/rooms/{room_name}` without being redirected to their talk, bypassing the scoping restriction intended for single-talk credentials.

app/routes/studio.py
```diff
@@ -415,6 +464,12 @@ def dashboard(
         all_talks = []
 
     all_rooms = sorted({t.room for t in all_talks if t.room})
+    # Stats reflect the scoped view, so event and room pages count only their talks.
+    if event_id is not None:
+        all_talks = [t for t in all_talks if t.event_id == event_id]
+    if room is not None:
+        all_talks = [t for t in all_talks if t.room == room]
+    current_event = next((e for e in user_events if e.id == event_id), None)
```
ViRUS-0-0: In `_render_talks_page`, `all_rooms = sorted({t.room for t in all_talks if t.room})` runs before `all_talks` is filtered by `event_id`. When an organizer who manages multiple conferences views a filtered event page (`/studio?event_id=...`), `all_rooms` retains rooms belonging to all of their events rather than just the active event. Any room dropdowns or UI controls referencing `all_rooms` display unrelated rooms.

Filter `all_talks` by `event_id` before computing `all_rooms`, ensuring room lists are strictly scoped to the active event.

app/ui/templates/dashboard.html.jinja
```diff
@@ -62,7 +82,7 @@
 
   <!-- Filter Bar -->
   <div class="filter-bar">
-    <form method="get" action="/studio" class="filter-form">
+    <form method="get" action="{{ filter_action }}" class="filter-form">
```
ViRUS-0-0, line 88: Conditionally render a hidden `<input type="hidden" name="event_id" value="{{ event_id }}">` inside the form whenever `event_id` is present.

app/routes/studio.py
```diff
@@ -300,7 +300,52 @@ def dashboard(
...
 
-    # 2. Check for authenticated user or active SSO session in cookie
+    return _render_talks_page(
+        request,
+        db,
+        client,
+        event_id=event_id,
+        status_filter=status_filter,
+        q=q,
+    )
```
ViRUS-0-0, line 355: In `dashboard()`, `_render_talks_page` is invoked without passing `login_next`. In `_render_talks_page`, `login_next` defaults to static `"/studio"`. When an unauthenticated user accesses a parameterized dashboard URL like `/studio?event_id=42&status_filter=waiting_for_files`, they are redirected to `/login?next=%2Fstudio`. After authentication, all query parameters are lost and the user lands on the unfiltered dashboard.

Resolution: Compute `login_next` from the current request URL path and query string by default inside `_render_talks_page` when not explicitly supplied.

tests/test_ui_events.py
```diff
@@ -719,3 +719,275 @@ def test_create_quick_talk_unauthorized_event(client: TestClient, db_session):
...
+            "email": org.email,
+            "password": "testpass123",
+            "next": f"/studio/rooms/Main%20Stage?event_id={event.id}",
+        },
+        follow_redirects=False,
+    )
+    assert resp.status_code in (302, 303)
+    assert resp.headers["location"] == (
+        f"/studio/rooms/Main%20Stage?event_id={event.id}"
+    )
```
ViRUS-0-0: In `test_login_next_returns_to_room_page`, `login_redirect` receives `Location: /login?next=/studio/rooms/Main%2520Stage%3Fevent_id%3D...`. However, the subsequent POST to `/login` does not use the `next` value received from the redirect or from the rendered login page. Instead, it manually injects a hardcoded payload `"next": f"/studio/rooms/Main%20Stage?event_id={event.id}"`. Furthermore, `test_room_page_requires_authentication` explicitly asserts the corrupted `%2520` string in `resp.headers['location']`. This test suite masked the double-encoding bug.

Resolution: Parse the `next` query parameter directly from `login_redirect`, submit it in the `/login` POST request, follow the redirect, and assert that the redirected page returns HTTP 200 with the room's talks.

app/ui/templates/dashboard.html.jinja
```diff
@@ -178,9 +206,13 @@
           {% endif %}
           <td class="td-title">
             <strong>{{ talk.title }}</strong>
-            <small>ID #{{ talk.id }} &bull; Event #{{ talk.event_id }}</small>
+            <small>ID #{{ talk.id }} &bull; <a href="/studio?event_id={{ talk.event_id }}" class="talk-meta-link talk-event-link">{{ talk.event.name if talk.event else ('Event #' ~ talk.event_id) }}</a></small>
```
ViRUS-0-0: In `dashboard.html.jinja` line 209 and `studio.html.jinja` line 38, links are generated with `href="/studio?event_id={{ talk.event_id }}"`. If a talk has a `None` or missing `event_id`, this renders `href="/studio?event_id="` or `href="/studio?event_id=None"`. Accessing `/studio?event_id=None` causes FastAPI query parameter validation to fail with an HTTP 422 error because `event_id` is typed as `int | None`.

Resolution: Render the event link conditionally only when `talk.event_id` is present; render plain text fallback if `talk.event_id` is null.

ViRUS-0-0 (changes requested):
The PR is not ready to be merged yet. I have left some comments. Please resolve them.

Here's a checklist to check if all the comments are resolved:
- [ ] Fix double URL-encoding of `login_next` in `app/routes/studio.py` (lines 330-332 and 361)
- [ ] Move talk-scoped SSO session check into `_render_talks_page` to prevent unauthorized room browsing in `app/routes/studio.py` (line 345)
- [ ] Preserve query parameters in `login_next` for default dashboard access in `app/routes/studio.py` (lines 303-310)
- [ ] Filter `all_talks` by `event_id` before computing `all_rooms` in `app/routes/studio.py` (lines 464-468)
- [ ] Add hidden `<input type="hidden" name="event_id" value="{{ event_id }}">` inside filter form in `app/ui/templates/dashboard.html.jinja` (line 85)
- [ ] Guard against nullable `talk.event_id` in `app/ui/templates/dashboard.html.jinja` (line 209) and `app/ui/templates/studio.html.jinja` (line 38)
- [ ] Fix `test_login_next_returns_to_room_page` and assertion in `tests/test_ui_events.py` (lines 869, 895) to assert and follow actual uncorrupted redirect targets

app/routes/studio.py
sanketkokate712: Fixed in e1f0193. `room_talks` no longer builds `login_next`. `_render_talks_page` now takes the decoded `request.url.path` plus the query string and encodes it exactly once when building the redirect. `/studio/rooms/Main Stage?event_id=5` now redirects to `/login?next=/studio/rooms/Main%20Stage%3Fevent_id%3D5`.

app/routes/studio.py
sanketkokate712: This check is already inside `_render_talks_page()`: the `if sso_user["scope_type"] == "talk"` redirect sits right after the auth gate, so `/studio/rooms/...` goes through it the same way `/studio` does. `dashboard()` only keeps the `?sso_token=` landing step. I added `test_room_page_redirects_talk_scoped_sso_to_its_talk` to prove it: a talk-scoped SSO session hitting a room page (with or without `event_id`) gets a 303 to its own talk.

app/routes/studio.py
sanketkokate712: Fixed in e1f0193. The room list now comes from a `SELECT DISTINCT room` query over the caller's scope, narrowed to the selected `event_id`. On `/studio?event_id=...`, the Attach Room Video list only offers that event's rooms. Covered by `test_event_page_room_list_is_scoped_to_event`.

app/ui/templates/dashboard.html.jinja
sanketkokate712: That hidden input is already in the form at `dashboard.html.jinja:110-111` (`{% if event_id %}<input type="hidden" name="event_id" value="{{ event_id }}">{% endif %}`), just below the status select. `test_room_page_lists_only_talks_in_that_room` asserts it renders. Submitting a filter on an event or room page keeps `event_id`, and Clear keeps it too.

app/routes/studio.py
sanketkokate712: Fixed in e1f0193. `login_next` is no longer a parameter. `_render_talks_page` builds it from the current path and query for every talks page, so `/studio?event_id=42&status_filter=waiting_for_files` returns to the same filtered view after login. Plain `/studio` still gives `/login?next=/studio`, so #240's tests are unchanged. Covered by `test_login_next_keeps_dashboard_filters`.

tests/test_ui_events.py
sanketkokate712: Fixed in e1f0193. The login tests now share a `_login_via_redirect` helper that takes `next` from the actual `Location` header, loads the login page, posts that same `next`, and follows the redirect. They assert a 200 on the room page with the right talks (and, for the dashboard, that `event_id` and `status_filter` are kept). `test_room_page_requires_authentication` now asserts the single-encoded `Main%20Stage`.

app/ui/templates/dashboard.html.jinja
sanketkokate712: Done in e1f0193. The dashboard renders the event and room links only when `talk.event_id is not none` and otherwise falls back to plain text. The studio breadcrumb falls back to plain text the same way. For what it's worth, `talks.event_id` is `nullable=False` in the model, so this is defensive.

sanketkokate712:
@SxxAq thanks, I've addressed both in e1f0193.

**1. Stats in SQL:** `all_talks` is gone. `_render_talks_page` builds one scope expression for the caller (SSO event, the organizer's or admin's own events, or the API client's events, and `false()` otherwise), then narrows it to `event_id` when one is set. From that scope:
- the stat cards come from `SELECT status, count(id) ... GROUP BY status`, with the room filter added on room pages
- the Attach Room Video list comes from `SELECT DISTINCT room`

No talk rows are loaded to compute either one.

**2. Double encoding:** `room_talks` no longer pre-encodes the path. `_render_talks_page` takes the decoded path plus the query string and quotes it once, so the redirect is `/login?next=/studio/rooms/Main%20Stage%3Fevent_id%3D5`. The login tests now follow the real `Location` header through to the room page.

app/routes/studio.py
```diff
@@ -312,8 +355,13 @@ def dashboard(
         )
 
     if not user and not sso_user and client is None:
+        # Send the user back to this exact page (path + filters) after login.
+        # request.url.path is already decoded, so the value is encoded once here.
+        login_next = request.url.path
+        if request.url.query:
+            login_next += f"?{request.url.query}"
```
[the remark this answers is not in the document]
sanketkokate712: Good catch, fixed in 032b685. It was worse than the filter form: Starlette builds `request.url` by re-parsing the decoded path, so for `/studio/rooms/Room%20%235` both `request.url.path` and `request.url.query` came back truncated. That broke the login `next` target as well.

What changed:
- The path and query now come from the ASGI scope (`scope[path]`, `scope[query_string]`), which keeps `?` and `#` intact. Routing already used it.
- `filter_action` is `quote(path, safe=/)`, so the form posts back to `/studio/rooms/Ask%20Me%3F%20Anything`.
- For `login_next`, I didn't switch to `raw_path`. Doing that would bring back `Main%2520Stage` for ordinary names, which the maintainers asked us to remove in this PR. Instead, only `%`, `?` and `#` are re-escaped before the single `quote()` for the `next` value. `Main Stage` still gives `next=/studio/rooms/Main%20Stage`, and `Room #5` gives `next=/studio/rooms/Room%20%25235`, which login decodes back to `/studio/rooms/Room%20%235`.

Tests (parametrized for `Ask Me? Anything` and `Room #5`):
- `test_room_page_keeps_question_mark_and_hash_in_room_names`: the dashboard link, room page heading, form action and a filtered request all stay on the right room
- `test_login_next_keeps_question_mark_and_hash_in_room_names`: checks the exact `Location`, then logs in with the `next` from that redirect and lands on the room with its talks

app/routes/studio.py
```diff
@@ -315,22 +331,63 @@ def dashboard(
...
+    """Resolve an event_id given as a numeric id or as an external_id/slug."""
+    if event_id is None:
+        return None
+    if isinstance(event_id, int):
+        return event_id
+    if str(event_id).isdigit():
+        return int(event_id)
+    ev = (
+        db.query(models.Event).filter(models.Event.external_id == str(event_id)).first()
+    )
```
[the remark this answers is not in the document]
sanketkokate712: Good catch, fixed in the latest commit. `uq_events_source_external_id` makes `external_id` unique only per source, so a global `first()` could pick another source's event and then the scope filter would render an empty page for an event the caller owns.

`_resolve_event_id` now takes the caller's events and matches the slug against those first, falling back to the global lookup (and `-1` for an unknown slug) so behaviour from #244 is otherwise unchanged. Resolution moved into `_render_talks_page` right after `user_events` is built, which is where the caller's scope is known, so both `/studio` and `/studio/rooms/{room}` get it. The SSO branch is untouched: it still pins `event_id` to the token's scope.

New test `test_event_slug_resolves_within_callers_own_events`: two organizers own events from different sources that share a slug, with the foreign one created first so a global lookup finds it first. Each organizer resolves the slug to their own event on both the dashboard and the room page. I checked the test fails against the previous resolver.

app/routes/studio.py
```diff
@@ -315,22 +331,73 @@ def dashboard(
...
+    if isinstance(event_id, int):
+        return event_id
+    slug = str(event_id)
+    if slug.isdigit():
+        return int(slug)
+    for event in scoped_events:
+        if event.external_id == slug:
+            return event.id
+    ev = db.query(models.Event).filter(models.Event.external_id == slug).first()
+    return ev.id if ev else -1
```
SxxAq: `Event.external_id` is only guaranteed to be unique per source, not globally unique across all organizers and sources. Falling back to an unscoped `db.query(models.Event).filter(models.Event.external_id == slug).first()` across the entire database means that if a slug is not found in `scoped_events`, it will resolve to an arbitrary event belonging to another organizer.
 
 For organizers and API key clients, `scoped_events` already represents all events the caller is authorized to view. If the slug does not match any event in `scoped_events`, it should resolve directly to `-1` (unauthorized / not found) rather than performing an unscoped global lookup.

app/routes/studio.py
```diff
@@ -342,8 +409,15 @@ def dashboard(
...
+        # Send the user back to this exact page (path + filters) after login.
+        # Only path delimiters are re-escaped so the whole target is encoded
+        # once as the `next` value (spaces become %20, not %2520).
+        path, query = _request_path_and_query(request)
+        login_next = path.translate(_PATH_DELIMITERS)
+        if query:
+            login_next += f"?{query}"
         resp = RedirectResponse(
-            url="/login?next=/studio",
+            url=f"/login?next={urllib.parse.quote(login_next, safe='/')}",
```
SxxAq, line 429: Pre-escaping delimiters with `_PATH_DELIMITERS = str.maketrans({"%": "%25", "?": "%3F", "#": "%23"})` before passing `login_next` to `urllib.parse.quote(..., safe='/')` causes double-percent-encoding because `%` is not in `safe='/'`. 
 
> For example, a `?` translated to `%3F` is escaped again by `quote` into `%253F` (and `#` into `%2523`). When the browser posts the form back on login, standard form submission passes this value through, leaving `%3F` in the path parameter instead of decoding back to `?` (which is why `_login_via_redirect` in `tests/test_ui_events.py` had to call `urllib.parse.unquote` manually to make the assertion pass).
 

 Instead of manually pre-translating characters, encode the target URL once using appropriate safe characters (e.g. `quote(target, safe="/?&=#")` or quote the path and query string with standard URL quoting rules).

ref L-37

SxxAq (changes requested):

app/routes/studio.py
sanketkokate712: Agreed, done in e0ce3cc. `_resolve_event_id` no longer takes `db` and no longer runs the global query:

```python
return next((e.id for e in scoped_events if e.external_id == slug), -1)
```

A slug that matches none of the caller's events resolves to `-1`, so it matches no talk. `scoped_events` is the organizer's or admin's own events, or the API client's events; SSO sessions never reach this path because their `event_id` is pinned to the token scope.

The rendered output was already scoped (an out-of-scope id was filtered out downstream and gave 0 results), so this isn't a disclosure fix, but the unscoped lookup was misleading and an extra query for nothing.

New test `test_event_slug_owned_by_another_organizer_resolves_to_nothing`: an organizer requesting another organizer's slug gets 0 results, no event heading and no event name. `test_event_slug_resolves_within_callers_own_events` still covers two organizers with same-slug events from different sources each landing on their own.

app/routes/studio.py
sanketkokate712: I'd like to keep this one as it is, because `%253F` is the correct encoding here and `safe="/?&=#"` breaks the room name. Sorry for the wall of text, but the two encoding layers are easy to talk past.

The room name really does contain a `?`. In a URL path that character has to stay percent-encoded as `%3F`, otherwise it starts the query string. So the path we want the user to land on is:

```
/studio/rooms/Ask%20Me%3F%20Anything?event_id=5
```

That whole string is then the *value* of the `next` query parameter, so its `%` has to be escaped once more. `%3F` becoming `%253F` in the `Location` header is that second layer, not a bug. One decode by the browser/form gives back `…Ask Me%3F Anything?event_id=5`, which is the correct path, and `%3F` is meant to survive that decode.

Traced through both versions:

```
current (escape %, ? and # in the path, then quote once)
  Location : /login?next=/studio/rooms/Ask%20Me%253F%20Anything%3Fevent_id%3D5
  login got: /studio/rooms/Ask Me%3F Anything?event_id=5
  -> path  : '/studio/rooms/Ask Me? Anything'   query: 'event_id=5'   ✅

suggested safe="/?&=#"
  Location : /login?next=/studio/rooms/Ask%20Me?%20Anything?event_id=5
  login got: /studio/rooms/Ask Me? Anything?event_id=5
  -> path  : '/studio/rooms/Ask Me'             query: ' Anything?event_id=5'   ❌
```

With `?` in `safe`, the unescaped `?` ends the path, so the user lands on a room called "Ask Me" and the rest becomes query junk. Same for `#`, which would turn into a fragment and never reach the server.

On the test: `_login_via_redirect` doesn't call `unquote`. It does

```python
next_target = urllib.parse.parse_qs(location.query)["next"][0]
```

which is a single decode of the query parameter, exactly what a browser does before the form posts the value back. It then posts that value and follows the redirect, asserting the final page is the right room. That's the round trip, not a workaround.

Why only `%`, `?` and `#` are escaped rather than quoting the whole path: quoting it in full would also encode the space, giving `Main%2520Stage` for ordinary names, which is the double-encoding @ViRUS-0-0 asked to remove earlier in this PR. Escaping just the three delimiters keeps normal names single-encoded (`next=/studio/rooms/Main%20Stage`) while names containing `?` or `#` still survive.

CodeRabbit raised the opposite concern on the same lines (that the delimiters weren't preserved) and confirmed this version handles it. Tests `test_room_page_keeps_question_mark_and_hash_in_room_names` and `test_login_next_keeps_question_mark_and_hash_in_room_names` cover `Ask Me? Anything` and `Room #5` through both the filter form and the login round trip.

Happy to change it if you'd rather avoid the nested encoding entirely — for example storing the target in a short-lived cookie instead of the `next` parameter — but I don't think `safe="/?&=#"` is the right fix.

app/routes/studio.py
```diff
@@ -388,7 +456,11 @@ def dashboard(
...
 
-        query = db.query(models.Talk).options(selectinload(models.Talk.jobs))
+        event_id = _resolve_event_id(event_id, user_events)
+
+        query = db.query(models.Talk).options(
+            selectinload(models.Talk.jobs), selectinload(models.Talk.event)
+        )
         if user:
             if user.role in ("organizer", "admin"):
                 org_event_ids = [e.id for e in user_events]
```
Rachit7168: You removed the `models.Talk.event_id` filtering from `query` in the `if user:` block, but it looks like you forgot to replace it. 

Because of this, `query` is no longer filtered by `event_id` or `org_event_ids` for organizers/admins. This means `query.all()` will fetch **all talks in the database** (data leak), and selecting an event from the UI won't actually filter the talks list (it will only filter the stats and rooms).

Since you elegantly compute a `scope` condition later on for the stats, you can fix this by simply applying that `scope` to `query` right before you fetch the talks:

```python
    # ... after computing `scope` and `stats_scope`:
    if event_id is not None:
        scope = and_(scope, models.Talk.event_id == event_id)
        
    # Apply the scope to the query to secure the data and apply filters
    query = query.filter(scope)

File: app/routes/studio.py [hunk 14]
```diff
@@ -411,6 +483,8 @@ def dashboard(
         else:
             query = query.filter(models.Talk.id == -1)
```
Rachit7168: When a user submits an HTML form with `method="get"`, the browser replaces the entire query string with the form's inputs. 

While your `filter_action` correctly preserves the room in the path, any existing `event_id` query parameter is lost when the user searches or filters by status. To fix this, you should include `event_id` as a hidden input so it is preserved on submission:

```jinja
    <form method="get" action="{{ filter_action }}" class="filter-form">
      {% if event_id %}
        <input type="hidden" name="event_id" value="{{ event_id }}">
      {% endif %}
      <!-- ... rest of form inputs ... -->

File: app/routes/studio.py [hunk 15]
```diff
@@ -388,7 +456,11 @@ def dashboard(
         else:
             user_events = []
 
-        query = db.query(models.Talk).options(selectinload(models.Talk.jobs))
+        event_id = _resolve_event_id(event_id, user_events)
```
Rachit7168, line 462: By optimizing `_resolve_event_id` to strictly check against `user_events`, API clients (who authenticate with `client` instead of `user`) can no longer use slugs, because `user_events` will be an empty list for them. 

If an API client tries to pass an `event_id` slug to this endpoint, it will fail to resolve and return `-1`. In the old code, slugs were resolved globally from the DB, so it worked for clients. If API clients are still expected to use slugs here, you might need to query the DB for the slug or pass their allowed events into `scoped_events`.

Rachit7168 (changes requested):
Thanks for working on this. Please address below comments :

app/routes/studio.py
sanketkokate712: Thanks for looking closely at this. The talk query was still filtered, though. The if user: block kept query.filter(models.Talk.event_id.in_(org_event_ids)) and the event_id narrowing, which is why the existing scoping tests (test_dashboard_talks_scoped_to_organizers_events, test_room_page_is_scoped_to_callers_events) passed. But I agree it was hard to see with the list and the stats each building the scope separately, so I took your suggestion in ef8b9db.

There's now one scope (the caller's visible events, narrowed to event_id when set), and the talk query, room list and stats all use it. The per-role branches only decide which events the caller can see. A plain user role gets an empty list, which gives an empty IN, so they see nothing. That removed 35 lines.

I also added test_talk_list_is_scoped_for_every_caller_type. It checks the rendered talk list (not just the stats) for an organizer, another organizer, an admin, a plain user and an API client, including filtering on someone else's event. For what it's worth, it passes on both the old and the new code.

app/routes/studio.py
sanketkokate712: This one's already in place: the hidden input is inside the filter form at dashboard.html.jinja:110-112, right after the status select:


```
{% if event_id %}
  <input type="hidden" name="event_id" value="{{ event_id }}">
{% endif %}
```
So searching or filtering by status keeps event_id, and the Clear button keeps it too. test_room_page_lists_only_talks_in_that_room checks that the input renders, and test_room_page_supports_status_and_search_filters submits a filter and checks the results stay on that event's room.

app/routes/studio.py
sanketkokate712: Good thing to check, but API clients keep slug support. user_events isn't empty for them: when the caller is an API client, it's set to the key's events (Event.id.in_(client.event_ids)) before _resolve_event_id runs. So a client's slug resolves against its own events.

I added test_api_client_resolves_slugs_for_its_events (ef8b9db). An API key uses a slug for its event on both /studio and the room page and gets that event's talks. A slug for an event outside the key's event_ids resolves to nothing. That's the intended difference from the old global lookup: a key can't reach an event it isn't assigned to by guessing its slug. This test also passes on the code before this commit.

ViRUS-0-0 (changes requested):
Please Resolve Conflicts

sanketkokate712:
> Please Resolve Conflicts


working!

Saksham-Sirohi (commented):
Re-checked on this HEAD against my earlier REQUEST_CHANGES: unauthenticated `/studio/rooms/...` no longer serves talk data — `_render_talks_page` redirects to login when there is no user/SSO/client, and the talk/room/stats queries are driven by a single caller-scoped `scope`. Room route + event/room links look correct; SSO talk-scoped breadcrumbs stay non-clickable.

sanketkokate712:
> Re-checked on this HEAD against my earlier REQUEST_CHANGES: unauthenticated `/studio/rooms/...` no longer serves talk data — `_render_talks_page` redirects to login when there is no user/SSO/client, and the talk/room/stats queries are driven by a single caller-scoped `scope`. Room route + event/room links look correct; SSO talk-scoped breadcrumbs stay non-clickable.

Thanks for re-checking, and for the original catch — the auth gate came straight from #240 once it landed, and moving it into _render_talks_page is what made the room route inherit it.

Since that review the branch has been rebased twice onto dev. The one change worth flagging: #253 added the speaker role, whose rule (own talks, across any event) doesn't fit "the events you can access", so it's folded into the same single scope — speakers match on speaker_email, everyone else on their accessible events, with the event narrowing applied on top. The talk list, room list and stats all still come from that one scope, so the Room page treats speakers the same way the dashboard does. #268's admin breadcrumb and #263's Edit button are untouched.

Current head is 902dc64.

ViRUS-0-0 (approved):
LGTM. Some new additions can be done in follow-up. This is fine for the time being.

SxxAq (approved):
LGTM

PR #267, opened 2026-09-18
feat(retention): implement scheduled final-only sweep job

### LINKED ISSUE #203: [Phase 6] Scheduled Cleanup Job — Final-Only Sweep
## Description

Narrower than the original design: since `raw/` is now retained indefinitely and `preview/`/`cut/` are handled by the immediate hooks, the only thing left needing a time-delayed sweep is `final/`, N days after `done`.

### Technical Design

- Periodic RQ job (RQ Scheduler or cron-triggered enqueue) running on a fixed interval, calling `run_retention_sweep()`.
- Sweep queries talks in `done` whose `updated_at` (time of entering `done`) plus the resolved `final_retention_days` has elapsed, and for which `final/` still exists.
- Re-resolves `RetentionPolicy` per event on each run (no caching), so an organizer changing their override takes effect on the next sweep without redeploy.
- Idempotent: a missing `final/` path is treated as already-clean, not an error, so re-running the sweep is safe.
- `rejected` talks are excluded from this sweep entirely — they have no `final/` to delete (transcode never ran) and their cleanup already happened via item 3's rejection hook.

## Scope

- Periodic job registration.
- `run_retention_sweep()`: query `done` talks past their `final/` retention window and delete `final/` where eligible.
- Idempotency handling.
- Tests: a `done` talk past its 14-day default has `final/` removed; a `done` talk within the window is untouched; an event with an "indefinite" override never has `final/` swept; a `rejected` talk is never touched by the sweep.

## Out of Scope

- Any `raw/`, `preview/`, or `cut/` handling — fully out of this job's responsibility under the revised policy.
- Sweep-history visibility beyond basic logging (Phase 7 observability candidate).

## Outcome

`final/` storage is bounded automatically without manual intervention, while `raw/` remains available indefinitely as the source of truth for future on-demand regeneration via the `ApprovedCut` record.

## Acceptance Criteria

- A `done` talk older than its resolved `final_retention_days` has `final/` deleted by the next sweep run.
- A `done` talk within the window is left untouched.
- Running the sweep twice produces no errors on already-deleted `final/` paths.
- Per-event overrides, including "indefinite," are honored.

## Dependency

Depends on the retention policy configuration for policy values, and on `states.py` reliably setting `updated_at` on entry into `done`.

SxxAq:
## Description
Implements Phase 6 of the storage retention lifecycle ([Issue #203](https://github.com/fossasia/veditor/issues/203)).

While intermediate artifacts (`cut/`, `preview/`, `intro/`, `outro/`) are handled on demand by lifecycle hooks upon talk completion/rejection and `raw/` media is kept indefinitely as the source of truth, `final/` artifacts require a time-delayed scheduled sweep to prevent unbounded disk growth.

This PR adds:
1. An `updated_at` column to the `talks` table with Alembic migration `0011` (revising `0010`).
2. Automatic timestamp tracking in `states.advance()` to record the exact time when a talk transitions state (specifically upon entry into `done`).
3. `run_retention_sweep()` to query `done` talks past their resolved retention period and delete `{talk_id}/final` storage.
4. Dynamic per-event retention override re-resolution on every sweep without requiring a process restart.
5. Indefinite retention override support (`"final_retention_days": "indefinite"`) ensuring protected talks are never touched.
6. Safe idempotency handling: missing or already-cleaned `final/` paths produce no errors.
7. Periodic job registration (`register_periodic_retention_sweep()`) and manual enqueue helper (`enqueue_retention_sweep()`).
8. CLI command `veditor admin run-retention-sweep` supporting synchronous runs and `--enqueue` background dispatch.
9. `--with-scheduler` argument in `scripts/run_worker.py` to enable worker scheduler support.

## Changes
- **`migrations/versions/0011_add_talk_updated_at.py`**: Alembic migration adding `updated_at` to `talks`.
- **`app/models.py`**: Added `updated_at` column to `Talk` model.
- **`app/schemas.py`**: Added `updated_at` to `TalkRead` schema.
- **`app/states.py`**: Set `talk.updated_at = datetime.now(UTC)` on successful transitions in `advance()`.
- **`app/config.py`**: Added `retention_sweep_interval_seconds` setting (default 3600s).
- **`app/retention.py`**: Added `run_retention_sweep()`, `enqueue_retention_sweep()`, and `register_periodic_retention_sweep()`.
- **`app/tasks.py`**: Re-exported retention sweep functions for worker import.
- **`app/cli.py`**: Added `admin run-retention-sweep` CLI subcommand.
- **`scripts/run_worker.py`**: Added `--with-scheduler` argument.
- **Tests**:
  - `tests/test_retention.py`: Tested sweep past default window, within window, custom day overrides, `"indefinite"` override, `rejected` talk exclusion, idempotency on repeated runs, dynamic override updating, and RQ enqueue/scheduler helpers.
  - `tests/test_states.py`: Tested `advance()` updating `talk.updated_at`.
  - `tests/test_models.py`: Tested `Talk.updated_at` column persistence.
  - `tests/test_cli.py`: Tested `admin run-retention-sweep` sync and `--enqueue`.
  - `tests/test_worker.py`: Tested `--with-scheduler` worker flag.

## Verification
- `pytest tests/test_retention.py tests/test_states.py tests/test_models.py tests/test_cli.py tests/test_worker.py`: 78 passed.
- `ruff check .`: All checks passed.
- `ruff format --check .`: 84 files already formatted.

Closes #203

## Summary by Sourcery

Implement scheduled cleanup of expired final artifacts with state-aware retention tracking, configurable policies, and operational controls.

New Features:
- Add scheduled and manually triggerable retention sweeps that remove expired `final/` artifacts while preserving raw media and indefinitely retained talks.
- Expose retention sweep execution through the admin CLI and optionally enable periodic scheduling in workers.

Bug Fixes:
- Prevent waveform generation from failing when source media disappears during processing.

Enhancements:
- Track talk state-transition timestamps and final-artifact cleanup status to support retention decisions and idempotent sweeps.
- Resolve event-specific retention policies dynamically on each sweep, including indefinite retention overrides.

Tests:
- Add coverage for retention timing, overrides, cleanup idempotency, scheduling, CLI execution, state timestamps, worker scheduling, and missing waveform sources.

## Summary by CodeRabbit

- **New Features**
  - Added automated retention sweeps to remove expired talk artifacts while preserving content within its retention period.
  - Added an admin command to run sweeps immediately or enqueue them for background processing.
  - Added configurable periodic scheduling and optional worker scheduler support.
  - Talk responses now include last-updated and final-cleanup timestamps.
  - Retention sweeps safely skip already-cleaned talks and report processed items or job IDs.

- **Bug Fixes**
  - Valid talk state transitions now record a timezone-aware update timestamp.


ViRUS-0-0:
@SxxAq Tests are failing

app/retention.py
```diff
@@ -76,3 +76,161 @@ def get_retention(event: Any | None = None) -> RetentionPolicy:
...
+        ):
+            logger.info(
+                "Periodic retention sweep job already registered (%s)",
+                existing_job.id,
+            )
+            return existing_job
+    except Exception as exc:  # noqa: BLE001
+        logger.debug(
+            "Periodic retention sweep job not found or connection failed: %s", exc
+        )
```
ViRUS-0-0, line 232: In register_periodic_retention_sweep(), Job.fetch(RETENTION_SWEEP_JOB_ID) checks if existing_job.get_status() is in ('queued', 'scheduled', 'started'). If the sweep job previously completed (status 'finished' within result_ttl) or failed (status 'failed'), the conditional check evaluates to False. Execution then drops through to target_queue.enqueue_in(..., job_id=RETENTION_SWEEP_JOB_ID, ...). Because the job key already exists in Redis, RQ raises rq.exceptions.JobAlreadyExistsError. When run on worker startup via scripts/run_worker.py, this exception is caught and logged as a warning, permanently aborting the registration of the periodic retention sweep for the lifetime of that worker.

scripts/run_worker.py
```diff
@@ -26,7 +30,22 @@ def _run_single_worker(
     engine.dispose(close=False)
     redis_conn = redis.from_url(redis_url)
     worker = Worker(queues, connection=redis_conn, name=name)
-    worker.work(burst=burst)
+    if with_scheduler:
+        from rq import Queue
+
+        from app.retention import register_periodic_retention_sweep
```
ViRUS-0-0: Imports at top always

scripts/run_worker.py
```diff
@@ -26,7 +30,22 @@ def _run_single_worker(
...
+
+        scheduler_queue = Queue("light", connection=redis_conn)
+        try:
+            register_periodic_retention_sweep(queue=scheduler_queue)
+        except Exception as exc:  # noqa: BLE001
+            print(
+                f"Warning: Failed to register periodic retention sweep on worker startup: {exc}",
+                file=sys.stderr,
+            )
+        worker.work(burst=burst, with_scheduler=True)
```
ViRUS-0-0, line 48: In scripts/run_worker.py, _run_single_worker hardcodes scheduler_queue = Queue('light', connection=redis_conn) and registers the sweep on the 'light' queue. However, RQ's worker scheduler (worker.work(with_scheduler=True)) only inspects and promotes scheduled jobs for queues that worker.queues is actively monitoring. If an operator launches a worker with `python scripts/run_worker.py heavy --with-scheduler`, the worker's scheduler monitors only the 'heavy' queue registry while the job resides in the 'light' registry. As a result, the retention sweep is never promoted to the active queue and never runs.

app/retention.py
```diff
@@ -76,3 +76,161 @@ def get_retention(event: Any | None = None) -> RetentionPolicy:
...
+
+    now = datetime.now(UTC)
+    swept_talk_ids: list[int] = []
+
+    talks = (
+        db.query(Talk)
+        .options(joinedload(Talk.event))
+        .filter(Talk.status == "done")
+        .all()
+    )
```
ViRUS-0-0, line 123: run_retention_sweep queries all talks in 'done' status across the entire database via db.query(Talk).options(joinedload(Talk.event)).filter(Talk.status == 'done').all(). For every historical talk older than the retention threshold, it calls storage.list_keys(final_prefix). Because swept talks remain in 'done' status and their updated_at timestamp does not change, every subsequent hourly sweep will re-query and re-list storage for every talk ever completed in the system's history. Once a system accumulates thousands of talks, thousands of sequential remote storage calls (e.g. S3 ListObjectsV2) will exceed the 600-second job_timeout, causing the retention job to fail consistently.

tests/test_states.py
```diff
@@ -17,6 +17,15 @@ def test_advance_legal_transitions():
             assert talk.status == next_state
 
 
+def test_advance_updates_talk_updated_at():
+    talk = DummyTalk("uploading")
+    assert not hasattr(talk, "updated_at")
+    advance(talk, "done")
+    assert hasattr(talk, "updated_at")
+    assert talk.updated_at is not None
+    assert talk.updated_at.tzinfo is not None
```
ViRUS-0-0, line 26: In test_advance_updates_talk_updated_at(), the test initializes a talk in state 'uploading' and invokes advance(talk, 'done'). Transitioning directly from 'uploading' to 'done' violates the state machine invariants established in app/states.py (LEGAL_TRANSITIONS). While DummyTalk might lack transition validation in some mocks, testing an illegal transition obscures regressions and creates a fragile contract test.

Resolution: Use a valid predecessor state for 'done' (such as 'ready' or the state designated in LEGAL_TRANSITIONS) or test a legal transition step so that advance() executes without violating domain rules.

app/retention.py
```diff
@@ -76,3 +76,161 @@ def get_retention(event: Any | None = None) -> RetentionPolicy:
...
+            return existing_job
+    except Exception as exc:  # noqa: BLE001
+        logger.debug(
+            "Periodic retention sweep job not found or connection failed: %s", exc
+        )
+
+    repeat = Repeat(
+        times=sys.maxsize if hasattr(sys, "maxsize") else 1_000_000,
+        interval=interval,
+    )
```
ViRUS-0-0, line 241: In register_periodic_retention_sweep(), repeat is instantiated as Repeat(times=sys.maxsize if hasattr(sys, 'maxsize') else 1_000_000, interval=interval). In RQ, setting times=None denotes an indefinitely repeating job. Using a large integer relies on implementation-dependent integer serialization in Redis and counter decrements, rather than RQ's native infinite repetition convention.

Resolution: Pass times=None to Repeat to configure infinite recurrence idiomatic to RQ.

ViRUS-0-0 (changes requested):
Please fix the tests and take a look at the comments

scripts/run_worker.py
```diff
@@ -56,6 +77,11 @@ def main(argv: list[str] | None = None) -> None:
         default=1,
         help="Number of worker processes to spawn (default: 1)",
     )
+    parser.add_argument(
+        "--with-scheduler",
```
Rachit7168, line 81: You added the --with-scheduler flag to scripts/run_worker.py and implemented the logic to register the periodic job, but you didn't update docker-compose.yml to use it! Because the flag isn't being passed, the worker-light service will start without the scheduler, and the periodic retention sweep will never actually run in the Docker environment.

app/retention.py
```diff
@@ -76,3 +76,206 @@ def get_retention(event: Any | None = None) -> RetentionPolicy:
...
+            continue
+
+        retention_delta = timedelta(days=policy.final_retention_days)
+        if now < updated_at + retention_delta:
+            continue
+
+        final_prefix = f"{talk.id}/final"
+        talk_id = talk.id
+        try:
+            existing_keys = (
```
Rachit7168: If a storage backend (or a mock in tests) does not implement `list_keys`, `hasattr` returns `False` and `existing_keys` becomes `[]`.

Because of `if not existing_keys:`, the loop will mark the talk as cleaned and `continue`, **completely skipping** the `storage.delete(final_prefix)` call! This means files will never actually be deleted on backends lacking `list_keys`.

Since `storage.delete()` is idempotent, you should restructure it to always delete if `list_keys` isn't available:

```python
        if hasattr(storage, "list_keys"):
            existing_keys = storage.list_keys(final_prefix)
            if not existing_keys:
                talk.final_cleaned_at = now
                try:
                    db.flush()
                except SQLAlchemyError:
                    db.rollback()
                    raise
                continue
        
        # Always attempt deletion if list_keys is missing, or if keys exist
        storage.delete(final_prefix)
        talk.final_cleaned_at = now

File: app/models.py [hunk 9]
```diff
@@ -178,6 +178,16 @@ class Talk(Base):
     outro_source: Mapped[str | None] = mapped_column(String(50), nullable=True)
     custom_intro_path: Mapped[str | None] = mapped_column(Text, nullable=True)
     custom_outro_path: Mapped[str | None] = mapped_column(Text, nullable=True)
+    updated_at: Mapped[datetime] = mapped_column(
+        DateTime(timezone=True),
```
Rachit7168, line 182: You added `updated_at` without an `onupdate=func.now()` or `onupdate=lambda: datetime.now(UTC)` argument. It currently relies on being manually updated in `states.py` when the talk transitions states (e.g. to `done`). 

This means if a user updates the talk's metadata (like title or room) in the UI *after* the talk is marked `done`, the `updated_at` timestamp won't change, and the retention window will continue counting down from the time it finished processing. 

This is perfectly fine if your intention is for retention to strictly be based on the "time of completion", but I wanted to point it out in case you intended for *any* metadata edit to reset the retention clock!

Rachit7168 (changes requested):
Thanks for working on this. Please address :

ViRUS-0-0 (approved):
LGTM

ArnavBallinCode (approved):
LGTM

Saksham-Sirohi (approved):
Prior CR items look addressed on this HEAD: `final_cleaned_at` avoids re-scanning cleaned talks, missing-`list_keys` still deletes, `--with-scheduler` is wired in docker-compose, and worker registration uses the light queue the scheduler owns. Sweep behavior and tests look solid

